PR #30, opened 2016-06-22
Update checked_array_typechecking.c

bryongloden:
On line 1198 'r' is an uninitialized variable, which is an error.

Found by https://github.com/bryongloden/cppcheck


dtarditi:
This line of code isn't supposed to type check.  It is expected to be rejected by the compiler.  Multiplying a pointer by a pointer is an illegal operation.

bryongloden:
@mjatallah <3

PR #31, opened 2016-06-28
Add feature tests for parsing bounds expressions in function parameter lists

dtarditi:
This change adds feature tests for parsing bounds expressions in function parameter lists.  These tests are for the checkedc-clang pull request [Extend clang IR with bounds expressions and parse bounds expressions for parameters](https://github.com/Microsoft/checkedc-clang/pull/8).

### DIFF
--- a/tests/bounds_declaration_parsing.c
+++ b/tests/bounds_declaration_parsing.c
@@ -1,0 +1,103 @@
+// Feature tests of parsing new Checked C bounds declarations.
+//
+// The following lines are for the LLVM test harness:
+//
+// RUN: %clang_cc1 -verify -fcheckedc-extension %s
+
+// TODO: typecheck bounds expressions in parameter declarations in a 
+// scope with all of the parameters, not just the parameters already
+// declared.  
+//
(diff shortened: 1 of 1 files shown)


reubeno:
At some point I assume we'll want additional test cases, e.g.:

- Bounds exprs on parameters in function declarations (vs. func definitions)
- Bounds exprs on non-checked pointer types?  On non-pointer types?
- Negative test cases for the classes of expressions not allowable in bounds exprs?
- Other possible parsing ambiguities (e.g. use of ternary conditional operator in bounds expr body)?

Should we add some of these?  I know you've only gotten as far as parsing but not to type-checking bounds exprs, so I'd understand if you think it makes sense to delay some of these additional cases until then...

dtarditi:
I'll commit this change and then add test cases for function declarations and other possible parsing ambiguities.  We can add those tests now.  The code should work.  

The clang implementation doesn't have the checks for expressions not allowable ain bounds exprs, so those tests need to wait.   Bounds declarations for unchecked pointer types and non-pointer types are more advanced features that I'm putting off for now in the clang implementation.

PR #32, opened 2016-07-09
Test parsing bounds expressions in parameter list scopes. 

dtarditi:
Bounds expressions for parameters can use parameters variables declared
at any point in the parameter list.  This means that bounds expressions must
be checked in a scope that contains all the parameters for the
parameter list.

- This change adds tests for that, by having bounds expressions for parameters
use parameters declared later in the parameter list.
- It also adds tests that subexpressions of bounds expressions can be
parenthesized expressions.
- Finally, this change adds tests that have misspelled contextual keywords
in bounds expressions.  These expressions should be rejected by a compiler.
These tests can be used also to test error messages and error recovery in
compilers for that situation.

### DIFF
--- a/tests/bounds_declaration_parsing.c
+++ b/tests/bounds_declaration_parsing.c
@@ -1,0 +1,103 @@
+// Feature tests of parsing new Checked C bounds declarations.
+//
+// The following lines are for the LLVM test harness:
+//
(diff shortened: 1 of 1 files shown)


awruef:
Looks good to me.

reubeno:
Looks good to me.

PR #33, opened 2016-07-14
Add feature tests for parsing bounds declarations.

dtarditi:
This change adds:
- Feature tests for parsing variable declarations with bounds
  declarations.  Most of the tests are for local variable declarations,
  with the assumption that parsing of bounds declarations themselves will
  be similar for variables with different linkage and storage classes.
  The tests also include tests for global variables with external and
  internal linkage.
- Feature tests for parsing function pointers where there are bounds on
  function parameters.

It also moves files to subdirectories so that tests will be easier to find.
The subdirectories are named after the functionality that is being tested
(parsing and typechecking, so far).   With the prior naming scheme, 
functionality was included at the end of the file name, making it harder to find 
tests.


reubeno:
Out of curiosity, would it be hard to break the tests up into separate sub-directories?  Or is it easiest for all of the tests to live in the same dir?  At some point you might want to use dir structure to group/partition/organize the tests.  For example, for now, you could have a parsin/ dir and typechecking/ dir (and remove that prefix from the filenames).

tests/parsing_bounds_parameters.c
```diff
@@ -168,3 +168,33 @@ extern void f32(int len,
                 int arr : count((len)),
                 int flag) {
 }
+
+// Pointers to functions that have bounds declarations on parameters.
```
reubeno: >functions [](start = 15, length = 9)

Would it also be useful to add some test cases for use of bounds expressions attached to the parameters of function types included in typedefs?  Is that even allowed?  For some reason, I vaguely recall previously having difficulty round-tripping parameter names in function typedefs with full fidelity; but it's also quite possible that it was user error.

tests/parsing_bounds_var_declarations.c
```diff
@@ -0,0 +1,141 @@
...
+extern array_ptr<int> h : bounds(f - 2, f + 3);
+extern array_ptr<int> i : count(5), j : bounds(i - 2, i + 3),
+                      k : bounds(j + 2, j + 5);
+
+array_ptr<int> b : count(3 + 2) = 0;
+array_ptr<int> d : byte_count(5 * sizeof(int)) = 0;
+array_ptr<int> g : bounds(none) = 0;
+
+static float vals[5] = { 0.0, 1.0, 2.0, 3.0, 4.0 };
+static array_ptr<float> l : count(5) = vals;
```
reubeno: >static [](start = 0, length = 6)

I can't imagine the implementation would care, but is it worth adding a test case for function-local statics?  Or any other storage classes?

reubeno:
I've left a few suggestions for additional cases and dir structuring, but it otherwise looks good to me!

tests/parsing_bounds_parameters.c
dtarditi: It is easy to move tests to subdirectories using the LLVM/clang test harness.  It recursively looks for tests in subdirectories of test directories.  I'll move the tests into separate sub-directories.

I'll add a few tests for function types in typedefs that have bounds declaration for parameters.  Yes, that is allowed.   C does have the restriction that `function type' for an identifier can't be achieved via a typedef.  You can use the typedef'ed function type to declare an identifier that has type 'pointer to function'.

tests/parsing_bounds_var_declarations.c
dtarditi: I added some feature tests for the other storage classes available in C,    I only added a few tests because that should be sufficient for testing most implementations.

dtarditi:
I've pushed commits that address your suggestions.

PR #35, opened 2016-07-20
Add feature tests for parsing return bounds expressions.

dtarditi:
This change adds tests for parsing return bounds expressions in function
declarators and function types.
- The tests for function declarators that return array_ptr types are
  straightforward.  array_ptr types avoid the problems of the
  infix C type declarator syntax, at least for pointer types.
- There are additional tests for parsing return bounds expressions for
  more complicated types, such as pointers to function types, function
  types that return pointers to function types, and function declarations
  that return  pointers to function types.
- There are variants of these additional tests for unchecked pointer types 
   with bounds declared for them.  The interoperation support in Checked C
   may lead to these types of bounds declarations.  The tests cover return 
   bounds expressions  for unchecked pointer types to function types and
   so on.

### DIFF
--- a/tests/parsing/return_bounds.c
+++ b/tests/parsing/return_bounds.c
@@ -5,8 +5,29 @@
 //
 // RUN: %clang_cc1 -verify -fcheckedc-extension %s
 
+// Parsing of function declarations
 extern array_ptr<void> alloc(unsigned size) : byte_count(size);
+extern array_ptr<int> f2(array_ptr<int> arr : count(5)) : count(3 + 2);
+extern array_ptr<int> f3(int len, array_ptr<int> arr : count(len)) 
+                      : count(len);
+extern array_ptr<int> f4(array_ptr<int> arr : byte_count(20))
+                      : byte_count(20);
+extern array_ptr<int> f5(array_ptr<int> arr : byte_count(5 * sizeof(int)))
+                      : byte_count(5 * sizeof(int));
+extern array_ptr<int> f6(array_ptr<int> arr : bounds(arr, arr + 5))
+                      : bounds(arr, arr + 5);
+extern array_ptr<int> f7(int start,
+                         array_ptr<int> arr : bounds(arr - start, arr - start + 5))
+                      : bounds(arr - start, arr - start + 5);
+extern array_ptr<char> f8() : bounds(none);
+// count, bounds, and none are contextual keywords.  They are treated as keyword
+// only when they immediately follow a ':';
+extern array_ptr<char> f9(int count) : count(count);
+extern array_ptr<char> f10(int none) : count(none);
+extern array_ptr<int> f11(int bounds, array_ptr<int> arr : count(bounds))
+                      : bounds(arr, arr + bounds);
 
+// Parsing function definitions.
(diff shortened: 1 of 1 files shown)


tests/parsing/return_bounds.c
```diff
@@ -0,0 +1,204 @@
...
+typedef int func2(array_ptr<int> arr : count(len), int len)
+            : bounds(arr, arr + len);
+static func1 *func1_ptr1;
+static func2 *func2_ptr1;
+func1 *func1_ptr2;
+
+extern void f25() : 6 + 6 { // expected-error {{expected bounds expression}}
+}
+
+extern array_ptr<char> f26() : count(len) { // expected-error {{use of undeclared identifier 'len'}}
```
reubeno: >count(len) [](start = 31, length = 10)

Should we have a version where there are no parameters but there's still a bounds expression (say a count(<some_const>).

While I'm thinking of it, we don't allow a reference to globals in bounds exprs, right?  Are we checking that now?

reubeno: Do we have to test with old-style function parameter lists where they go after the parentheses?  e.g.:

```C
void foo()
    int x;
{
    ....
}
```

Do we need/want to support code bases that still use them?  Will they interfere with our colon syntax?

tests/parsing/return_bounds.c
```diff
@@ -0,0 +1,204 @@
+// Feature tests of parsing new Checked C bounds declarations for
+// function return values
+//
+// The following lines are for the LLVM test harness:
+//
+// RUN: %clang_cc1 -verify -fcheckedc-extension %s
+
+extern array_ptr<void> alloc(unsigned size) : byte_count(size);
```
reubeno: I'd imagine it's fine that most of the cases in this file are against function definitions and not declarations.  Are there any reasons we should have more declaration-only cases?

For example, I know that these two declarations are different in C (not C++), and the first is specific to declarations (vs. definitions):

```C
int foo();
int foo(void);
```

The first declaration above indicates nothing about the parameters of the declaration; you can technically pass any number of arguments into it, if I recall correctly.  The second declaration provides a prototype--and explicitly indicates that it has 0 parameters.

Perhaps we should test both forms?

reubeno:
The test cases you have look pretty good; I've added a few suggestions that came to mind for additional edge test cases to consider adding...

tests/parsing/return_bounds.c
dtarditi: That's a good suggestion to add test cases where the argument expressions to count and bounds operators are omitted.    I'll add those cases.

With regard to old-style C function declarations, we first need to decide on the language design first.  I've opened issue #36 for this.  We can add modify clang and add tests once we make those decisions.

We are not checking lifetime requirements on variables used in bounds expressions yet.   We will allow references to globals in some cases.  For example, it might make sense for a local variable to have bounds that reference a global buffer.

tests/parsing/return_bounds.c
dtarditi: I agree that it makes sense to add more tests of declarations of functions.

For `int foo()', you are correct.   We need to decide on the language design (issue #36) first.

PR #37, opened 2016-07-29
Feature tests for parsing member bounds.

dtarditi:
This change adds feature tests for parsing member bounds.   This accompanies the pull request 23 for the checked-clang repo.

### DIFF
--- a/tests/parsing/member_bounds.c
+++ b/tests/parsing/member_bounds.c
@@ -110,7 +110,7 @@ struct S14 {
   // Members that are array_ptrs to checked arrays
   array_ptr<int checked[10]> arr7 : count(5);
   array_ptr<int checked[10]> arr8 : byte_count(5 * sizeof(ptr<int>));
-  array_ptr<int checked[10]> arr9 : bounds(arr9, arr3 + 9);
+  array_ptr<int checked[10]> arr9 : bounds(arr9, arr9 + 9);
   // Members that are unchecked pointers with bounds.  These will
   // be used for interoperation.
   int *arr10 : count(5);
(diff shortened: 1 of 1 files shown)


tests/parsing/member_bounds.c
```diff
@@ -0,0 +1,248 @@
...
+  array_ptr<int checked[10]> arr7 : count(5);
+  array_ptr<int checked[10]> arr8 : byte_count(5 * sizeof(ptr<int>));
+  array_ptr<int checked[10]> arr9 : bounds(arr9, arr3 + 9);
+  // Members that are unchecked pointers with bounds.  These will
+  // be used for interoperation.
+  int *arr10 : count(5);
+  int *arr11 : byte_count(5 * sizeof(int));
+  int *arr12 : bounds(arr12, arr12 + 9);
+  // Members that are arrays and that have bounds specified for them.
+  // Legal but not too interestting.
```
reubeno: >t [](start = 30, length = 1)

[Nit] Typo.

tests/parsing/member_bounds.c
```diff
@@ -0,0 +1,248 @@
...
+  // not a keyword
+  array_ptr<int> arr4 : bounds(arr3, arr3 + none); 
+};
+
+struct S13 {
+  // 'bounds' is a contextual keyword.  It is only a keyword when it follows
+  // the ':' in a bounds declaration.
+
+  // not a keyword
+  int bounds;
```
reubeno: >bounds [](start = 6, length = 6)

We chatted about this earlier, but it might be nice to have a test case like this too:

constexpr int bounds = 4;

struct S {
     int foo : bounds;
};

Regardless of whether we accept or reject this program, it would be nice to have a test case to clarify the expected results.

tests/parsing/member_bounds.c
```diff
@@ -0,0 +1,248 @@
+// Feature tests of parsing new Checked C member bounds declarations.
+//
+// The following lines are for the LLVM test harness:
+//
+// RUN: %clang_cc1 -verify -fcheckedc-extension %s
+
+struct S1 {
```
reubeno: >struct [](start = 0, length = 6)

I know the logic is pretty well shared, but should we also add some cases for unions?

reubeno:
C11 also supports anonymous struct and unions; could we add a couple of quick tests to cover them showing up as members in a struct, with bounds expressions referencing members inside the inner (anonymous) struct?

reubeno:
Otherwise, looks good :)

Issue #11, opened 2016-04-20
discuss terminology of safe vs. unsafe pointers


dtarditi:
We are using the terminology of safe and unsafe pointer types in the document.  It might be better to call them checked and unchecked pointer types instead.  There are fewer negative connotations to the term unchecked.

dtarditi:
The spec has been updated to incorporate the feedback.

PR #40, opened 2016-08-02
Update description of checked pointers.

dtarditi:
This finishes up addressing issue #11.  When we switched to the terminology of
checked and unchecked pointers, there was feedback that we should be very
clear about what is being checked.  Specifically, checked pointers check
bounds, but do not check memory lifetime.  This adds a paragraph to the
description of checked pointers making to clear that memory lifetime is
not checked.


awruef:
It makes sense, but it seems a little against the spirit of the document to refer to things as safe and checked in the same sentence. What about the following?

Checked pointers ensure the spatial validity of an access, that the access is within an explicitly provided bounds of an object. They do not the temporal validity of an access. It’s still possible to access a region of memory after it has been deallocated, also known as “use-after-free.”

awruef:
It's a very small nit but you could change the beginning of the last sentence to say "For example, they ..."

PR #46, opened 2016-08-23
Test checked properties of dimensions of multi-dimensional arrays

dtarditi:
This change adds feature tests that the dimensions of a multi-dimensional array are either all checked or unchecked.   The Checked C spec requires this for multi-dimensional arrays.     It removes feature tests where individual dimensions could be declared as unchecked using the `unchecked` keyword.   These feature tests were for an experimental feature in the clang implementation of Checked C.  We have decided to stick with the existing wording in the spec.

This change also adds tests for parenthesized array declarators.   The checked property is supposed to propagate from the outer dimension of a multi-dimensional array to the inner dimension: `int a checked[10][10]` declares a checked array of checked arrays  This propagation wasn't happening when the declarator for the outer dimension was parenthesized: `int (a checked[10])[10].`

### DIFF
--- a/tests/parsing/checked_array_types.c
+++ b/tests/parsing/checked_array_types.c
@@ -76,9 +76,6 @@ extern void k1(int y)
     int arr2 checked[5];
     int arr3 checked[][5] = { { 1 }, {2 } };
     int arr4 checked[5][5];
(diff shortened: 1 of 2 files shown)


reubeno:
Looks good to me.

PR #47, opened 2016-08-25
Test typechecking of count bounds expressions.

dtarditi:
Also test typechecking of byte_count bounds expressions.

### DIFF
--- a/tests/typechecking/bounds.c
+++ b/tests/typechecking/bounds.c
@@ -6,7 +6,11 @@
 
 // Test expressions with standard signed and unsigned integers types as
 // arguments to count and byte_count.
-extern void f1() {
+
+static int A = 8;
+static long long int B = 8;
+
+extern void f1(void) {
 
   char c1 = 8;
   short c2 = 8;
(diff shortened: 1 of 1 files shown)


tests/typechecking/bounds.c
```diff
@@ -0,0 +1,117 @@
...
+  long long int c5 = 8;
+
+  _Bool c6 = 1;
+  unsigned char c7 = 8;
+  unsigned short c8 = 8;
+  unsigned int c9 = 8;
+  unsigned long int c10 = 8;
+  unsigned long long int c11 = 8;
+
+  array_ptr<int> t1 : count(c1) = 0;
```
reubeno: >0 [](start = 34, length = 1)

Should we add some cases with negative integer literals?  With floating-point literals?  With other sorts of expressions on the right hand size of the '='?  Or do we already have those cases covered elsewhere?

tests/typechecking/bounds.c
```diff
@@ -0,0 +1,117 @@
...
+  array_ptr<int> t10 : byte_count(c1) = 0;   // expected-error {{invalid argument type 'float' to byte_count expression}}
+  array_ptr<int> t11 : byte_count(c2) = 0;  // expected-error {{invalid argument type}}
+  array_ptr<int> t12 : byte_count(c3) = 0;  // expected-error {{invalid argument type}}
+  array_ptr<int> t13 : byte_count(c4) = 0;  // expected-error {{invalid argument type}}
+  array_ptr<int> t14 : byte_count(c5) = 0;  // expected-error {{invalid argument type}}
+  array_ptr<int> t15 : byte_count(c6) = 0;  // expected-error {{invalid argument type}}
+  array_ptr<int> t16 : byte_count(c7) = 0;  // expected-error {{invalid argument type}}
+  array_ptr<int> t17 : byte_count(s.f) = 0;  // expected-error {{invalid argument type}}
+#ifndef __STDC_NO_COMPLEX__
+  array_ptr<int> t18 : byte_count(c8) = 0;   // expected-error {{invalid argument type}}
```
reubeno: >c8 [](start = 34, length = 2)

What about naming a function here?  Or a global/static?  Or non-symbol expressions?

dtarditi:
Thanks for the suggestions!   I've added some additional negative tests case, as well as a few tests cases involving the use of integer literals or file/globally scoped variables.

dtarditi:
Thanks for the feedback.  I've added some additional negative tests cases, as well as a few tests cases involving the use of integer literals or file/globally scoped variables.   

I don't think we need to try different right-hand sides - that doesn't really affect type checking of the count expressions.    Different right-hand sides are covered by other tests.   However, I did realize that we need to check the type of the variable that has the count bounds expression.   The spec requires additional checking on the type.  I've opened issue 41 in the Checked C clang repo for this.

dtarditi:
ReubenO OK'ed the pull request, so I've merged it in.

PR #48, opened 2016-08-31
Tests for checking type requirements for bounds declarations.

dtarditi:
There is a pull request for the checked-clang repo (https://github.com/Microsoft/checkedc-clang/pull/42) that adds checking of type requirements for bounds declarations to clang.  See the pull request for details. 

This commit adds the matching tests.   It includes both positive tests (that are supposed to work) and negative tests that should fail with an error message.    We have added the expected error message from the clang implementation for the negative tests, so that the tests can be used in the clang test infrastructure.  This commit also fixes issues in existing tests that the checking uncovered.

### DIFF
--- a/spec/bounds_safety/variable-bounds.tex
+++ b/spec/bounds_safety/variable-bounds.tex
@@ -75,6 +75,10 @@ In a bounds declaration of the form \boundsdecl{\var{x}}{\var{bounds-exp}},
 \var{x} must have an \arrayptr\ or a checked array type. 
 For the form  \boundsdecl{\var{x}}{\boundscount{\var{e1}}},  the type of 
 \var{x} cannot be \arrayptrvoid\ and the type of \var{e1} must be an integral type. 
+The usual C integer conversions are applied to \var{e1}\footnote{If the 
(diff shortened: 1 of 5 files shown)


spec/bounds_safety/variable-bounds.tex
```diff
@@ -72,14 +72,18 @@ \section{Bounds declarations}
...
+the expression is promoted to an \keyword{int} type if that is large
+enough to  hold all values of the type or an \keyword{unsigned int} type otherwise.}
 For \boundsdecl{\var{x}}{\bounds{\var{e1}}{\var{e2}}}, \var{e1} and
-\var{e2} must have the same \arrayptr\ type. The type of x is
-typically the same type as \var{e1} and \var{e2}, but it can be a
-different \arrayptr\ or checked array type. Allowing the types
-to be different is useful for describing the results of pointer casts and
+\var{e2} must be pointers to the same type.  Typically \var{x} is also 
+a pointer to that type or an array of that type.  
+However, \var{x} can point to or be an array of a different type.
```
awruef: Maybe \var{x} can be a pointer to or an array of?

reubeno:
Looks reasonable to me.

PR #52, opened 2016-09-15
Change Checked C keywords to avoid conflicts with existing identifiers.

dtarditi:
We're updating the clang implementation so that Checked C keywords start with an underscore and a capital letter. This avoids conflicts with existing C identifiers.

The keyword change follows the rules in the C Standard: all identifiers beginning with an underscore followed by a capital letter or underscore are reserved for implementation use.  This allows them to be used for language extensions without conflicting with existing identifiers. Of course, some programs do not follow the standard and use these identifiers.

This change adds a header file stdchecked.h that has macros that maps the existing non-prefix, lower case keywords to the new keywords.  It updates the tests to include that header file.

Testing:
- Updated version of clang passes these new tests.

### DIFF
--- a/tests/parsing/checked_array_types.c
+++ b/tests/parsing/checked_array_types.c
@@ -16,6 +16,8 @@
 // parameter have new checked array types
 //
 
+#include "../../include/stdchecked.h"
+
 extern void f1(int a checked[], int y) {
 }
 
--- a/include/stdchecked.h
+++ b/include/stdchecked.h
@@ -1,0 +1,10 @@
+#ifndef __STDCHECKED_H
+#define __STDCHECKED_H
+
+#define ptr _Ptr
+#define array_ptr _Array_ptr
+#define checked _Checked
+#define unchecked _Unchecked
+#define where _Where
+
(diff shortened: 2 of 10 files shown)


tests/parsing/checked_array_types.c
```diff
@@ -16,6 +16,8 @@
 // parameter have new checked array types
 //
 
+#include "../../include/stdchecked.h"
```
reubeno: Is the desire here to avoid churning the test code too much, or is it that you'd prefer to write the tests in terms of the convenience #defines?  In other words, when you author new test code, will you use _Ptr or ptr.  My recommendation is that we should use in the tests whatever it is we expect customers to use, use in documentation and samples, etc.

dtarditi: I expect customers to use the shorter names  For existing code, this may be a bit of an issue because including the header file will be problematic until system headers are cleaned up.   Generally, one expects to be able to include system header files in any order.  If a system header file has a use of ptr in it, that would be break if stdchecked.h is included first.

PR #56, opened 2016-09-17
Add tests of parsing interop type annotations.

dtarditi:
This change adds tests for parsing of bounds-safe interface type annotations.   It matches a corresponding change for the checked-clang repo. 

The type annotations are a generalization of the `int * x : ptr` annotation described in the specification. That annotation does not provide a way to represent information for pointers with multiple levels of indirection (`int **x` for example). Instead of just using `ptr`, we now allow a type annotation to be specified.  The annotation has the form `type` `(` _type name_ `)`.   For example, a variable declaration `int **x` could be annotated as `int ** x : type(_Ptr<_Ptr<int>>)`.  This richer form allows us to represent bounds-safe interfaces where pointers are passed by reference, for example.

Bounds-safe interfaces are allowed for parameters, global variables, and members.   This change adds tests for type annotations involving `_Ptr`, `_Array_ptr`, typedef'ed names, and types with type modifiers.

### DIFF
--- a/tests/parsing/interop_types.c
+++ b/tests/parsing/interop_types.c
@@ -1,0 +1,94 @@
+// Feature tests of parsing Checked C interop annotations that declare
+// an alternate checked pointer type for a variable or member with
+// unchecked pointer type.
+//
+// The following lines are for the LLVM test harness:
+//
+// RUN: %clang_cc1 -verify -fcheckedc-extension %s
+// expected-no-diagnostics
+
+#include "../../include/stdchecked.h"
+
+//
+// parameters with interop type annotations
+//
+
+// first parameter has interop type annotation
+extern void f1(int *p : ptr<int>, int y) {
+   *p = y;
+}
+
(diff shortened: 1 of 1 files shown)


reubeno:
I'm taking a look at this now; quick question, though: are we requiring that the bounds-safe type be equivalent to the original C type if we were to replace all checked types with their unchecked equivalents.  In other words, are we doing anything to prevent something like this?

`void some_function(int *p : _Ptr<`**`float`**`>, int count);`

Also, if we were to correct the type mismatch, where would the `count()` bounds expression go in the example above?  My apologies if this was all covered long ago; my memory's a bit foggy on some of this...  (IOW, feel free to tell me to read the darn manual.)

dtarditi:
I've revised the syntax for type annotations.

To answer Reuben's question, yes, we will require that the bounds-safe type be equivalent to the original C type if we replace checked types with their unchecked equivalents.  There will be a future PR where these type requirements are checked.

PR #59, opened 2016-10-03
Feature tests for bounds-safe interface implicit type conversions.

dtarditi:
This change adds feature tests for implicit type conversions at bounds-safe interfaces.   It includes three sets of tests, corresponding to each language construct that can have a bounds-safe interface: a set for parameters with bounds-safe interfaces, a set for non-locally scoped variables with bounds-safe interfaces, and a set for members with bounds-safe interfaces.  These tests match corresponding compiler changes for Checked C clang repo issue 31.

For each set of tests, there are
- tests for assigning (or passing) checked pointers to unchecked pointers with bounds-safe interfaces (where the referent types match)
- tests for assigning or passing checked pointers to void * pointers with bounds-safe interfaces
- tests for passing checked void pointers to non void * pointers with bounds-safe interfaces.  These tests are expected to cause errors.
- There are also tests for types involving type qualifiers, making sure that the qualifier restrictions are enforced (a pointer to  non-const type can be assigned to a pointer to a const-type, but not the reverse).

### DIFF
--- a/tests/typechecking/interop.c
+++ b/tests/typechecking/interop.c
@@ -1,4 +1,4 @@
-// Feature tests of typechecking pf uses of Checked C interoperation
+// Feature tests of typechecking of uses of Checked C interoperation
 // declarations.
 //
 // The following lines are for the LLVM test harness:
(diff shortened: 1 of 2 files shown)


tests/typechecking/interop.c
```diff
@@ -0,0 +1,394 @@
+// Feature tests of typechecking pf uses of Checked C interoperation
```
reubeno: Is `pf` a typo or standard abbreviation I'm not familiar with?

tests/typechecking/interop.c
```diff
@@ -0,0 +1,394 @@
...
+
+void f3(int *p : byte_count(len * sizeof(int)), int len) {
+}
+
+void f4(int *p : bounds(p, p + len), int len) {
+}
+
+// void * parameters with interop declarations.  Note that count bounds
+// expressions are not allowed for void * pointers because they don't
+// make sense.
```
reubeno: Good point; can we add a negative test case for it?

tests/typechecking/interop.c
```diff
@@ -0,0 +1,394 @@
...
+}
+
+void f3_const(const int *p : byte_count(len * sizeof(int)), int len) {
+}
+
+void f4_const(const int *p : bounds(p, p + len), int len) {
+}
+
+// Pointers to non-const qualified data can be passed to parameters that are pointers
+// to const-qualified ata.
```
reubeno: Typo: missing `d` from `data`?

tests/typechecking/interop.c
```diff
@@ -0,0 +1,394 @@
...
+#include "../../include/stdchecked.h"
+
+//
+//
+// Pass an argument with a checked pointer type to a parameter with
+// an unchecked pointer type and an interop bounds declaration.
+//
+//
+
+void f1(int *p : type(ptr<int>)) {
```
reubeno: I've forgotten; do we already have test cases that check whether an interop bounds declaration can be added to fields or parameters with non-pointer types?  For ex: a uintptr_t.

reubeno: How about array types?  Both complete and incomplete...

dtarditi:
Thanks for the review.  That is a great catch on array types.  I hadn't even considered the possibility of allowing interop types on unchecked array types.  That makes lots of sense, so I'll add the tests.

 `pf` is a typo.  I'll fix the other typos too.   I'll also add the negative test, just to make that point clear.

We don't have tests case yet restricting the types to which bounds annotations can be applied.   The functionality isn't there yet in the compiler.   I'll adding both the compiler functionality and the tests soon.

PR #67, opened 2016-10-31
Test redeclarations of no-prototype functions with checked types.

dtarditi:
Calls to no-prototype functions are not type checked according to the C Specification (Section 6.5.2.2 of the C11 specification).  Because of this, the Checked C extension does not allow no-prototype functions to be redeclared to take or return checked types.   The lack of type checking at calls enables checking of bounds to be bypassed. Section 5.5 of the Checked C extension specification describes the restrictions.  In the terminology of the C Specification, no-prototype function types are incompatible with function types that take arguments or return values that have checked types, contain values with checked types embedded within them, or are recursively pointers to function types that take or return checked types or checked values.

This change adds feature tests that check that redeclarations of no-prototype functions to use checked types are not allowed.  The tests are in  typechecking/no_prototype_functions.c.

It also fixes accidental declarations in existing tests of no-prototype functions.  All these functions actually take no arguments.  In C, a prototype for a function that takes no arguments has to be declared with an
argument list that is the `void` keyword.  For example, `int f(void);` declares a function that takes no arguments and returns an integer.

### DIFF
--- a/tests/parsing/checked_array_types.c
+++ b/tests/parsing/checked_array_types.c
@@ -165,7 +165,7 @@ typedef int incomplete_checked_array_ty checked[];
 // Operators that take types
 //
 
-void parse_operators_with_types() {
+void parse_operators_with_types(void) {
     int s1 = sizeof(int checked[10]);
     int s3 = sizeof(ptr<int checked[5]>);
     int s4 = sizeof(array_ptr<int checked[5]>);
--- a/tests/typechecking/no_prototype_functions.c
+++ b/tests/typechecking/no_prototype_functions.c
@@ -1,0 +1,207 @@
+// Feature tests of typechecking no prototype functions.
+//
+// The following lines are for the LLVM test harness:
+//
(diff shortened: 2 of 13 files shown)


tests/typechecking/no_prototype_functions.c
```diff
@@ -0,0 +1,268 @@
+// Feature tests of typechecking no prototype functions.
```
reubeno: Should we add a case for a `typedef` use of no-prototype functions with Checked C types?

tests/parsing/checked_array_types.c
```diff
@@ -165,7 +165,7 @@ typedef int incomplete_checked_array_ty checked[];
 // Operators that take types
 //
 
-void parse_operators_with_types() {
+void parse_operators_with_types(void) {
```
reubeno: Observation: the fact that our own tests needed to be so broadly updated for this should probably signal that we're going to encounter quite a bit of code that will need to do the same.  Sounds like a great opportunity for our automatic translation/annotation tool to auto-add missing `void`s.

Issue #53, opened 2016-09-16
Change specification to describe updated Checked C keywords


dtarditi:
We've changed the Checked C keywords to begin with an underscore and a capital letter.   This is to avoid conflicts with identifiers in existing programs.   The keywords are _Ptr, _Array_ptr, _Checked, _Unchecked, and _Where.  We will have a header file "stdchecked.h" that defines macros that map the original names to these new types.

We plan to use the original shorter names through the spec because that is the way we intend for code to be written.  We need to describe the actual keywords and the assumption of the use of a header file when we introduce the new pointer types.

dtarditi:
The spec has been updated.

PR #68, opened 2016-11-15
Change specification to describe updated Checked C keywords.

dtarditi:
This change addresses issue #53.  To avoid backward-compatiblity problems when using the
Checked C extension with existing C programs, we are changing the keywords for the
extension to start with an underscore followed by a capital letter.  

The updated keywords are:

``
_Array_ptr  _Checked  _Dynamic_check  _Ptr  _Span  _Where  _Unchecked 
``

This change adds a description of these keywords to the specification.  We use a header file 
that provides all-lowercase versions of the keywords.  The text and examples in the
specification continue to use the all-lowercase versions because that is the way extension is
intended to be used.

This change also clarifies that the syntax for bounds expressions uses specific identifiers,
not keywords.   This cuts down on the number of new keywords that need to be introduced.
The places where the specific identifiers occur in bounds expressions are not places where 
general-purpose identifiers can also appear.  This means there is no need to turn the specific
identifiers into keywords to disambiguate a possible overlap during parsing.  This technique
has been used in C++ and C# to avoid introducing new keywords for names that have special 
meaning only in specific syntactic contexts.

### DIFF
--- a/spec/bounds_safety/core-extensions.tex
+++ b/spec/bounds_safety/core-extensions.tex
@@ -13,6 +13,47 @@ proves them redundant) because bounds safety depends upon them.
 Finally, it describes changes to
 undefined behavior needed for bounds safety.
 
+\section{New keywords}
+Some of the extensions use new keywords.   Introducing new keywords
+into a programming language can cause conflicts with identifier names
+in existing programs, if the keywords can appear in places where
+identifiers could also appear.   Fortunately, C provids a
+backward-compatible way to introduce new keywords.  Identifiers that
+begin with an
+underscore followed by a capital letter are reserved for system
+use, as are identifiers that begin with two underscores 
+\cite[Section 7.1.3]{ISO2011}.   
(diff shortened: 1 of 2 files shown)


spec/bounds_safety/core-extensions.tex
```diff
@@ -13,6 +13,42 @@ \chapter{Extensions to existing C concepts}
...
+begin with two underscores \cite[Section 7.1.3]{ISO2011}. The following
+new keywords are introduced:
+\begin{verbatim}
+_Array_ptr  _Checked  _Dynamic_check  _Ptr  _Span  _Where  _Unchecked
+\end{verbatim}
+
+It is desirable to have all-lowercase versions of the
+identifiers for readability and ease of typing. The C
+preprocessor is used to provide these. A standard header
+file \keyword{checkedc.h} is provided that has macros
```
reubeno: Should we name it something like `stdcheckedc.h`?

PR #99, opened 2017-01-25
Adds Explicit Dynamic Check

lenary:
This adds the `dynamic_check` macro for `_Dynamic_check`, and some initial tests for its functionality

These initial tests make sure that successful programs don't exit with a failure on a dynamic check failure, and that dynamic checks that fail cause an exit with an error. Unfortunately, the execution of `llvm.trap()` counts as a crash for the "not" tool, which could make these tests more brittle. But they at least pass without having to add a SEH or similar.


tests/dynamic_checking/failing-arith-dynamic_check.c
```diff
@@ -0,0 +1,19 @@
...
+// RUN: not --crash %t.exe
+
+#include "../../include/stdchecked.h"
+
+void f1(int i) {
+  dynamic_check(i < 30);
+}
+
+int main(void) {
+  f1(50);
```
lenary: Done, added on the `dynamic_check` expression.

PR #102, opened 2017-01-27
Adds tests for function pointer casts

lenary:
These are the tests for Microsoft/checkedc-clang#125

### DIFF
--- a/tests/typechecking/function_casts.c
+++ b/tests/typechecking/function_casts.c
@@ -148,4 +148,14 @@ void local_convert(int(*f)(int)) {
   ptr<int(int)> bad_cast_2 = (ptr<int(int)>)(bad_f0); // expected-error {{cannot cast an unchecked function pointer to a checked pointer unless unchecked function is named}}
   ptr<int(int)> bad_cast_3 = (ptr<int(int)>)(*bad_f0); // expected-error {{cannot cast an unchecked function pointer to a checked pointer unless unchecked function is named}}
   ptr<int(int)> bad_cast_4 = (ptr<int(int)>)(***bad_f0); // expected-error {{cannot cast an unchecked function pointer to a checked pointer unless unchecked function is named}}
+
+  //
+  // Weird Casts
+  //
+
+  ptr<float(float, int)> local_safe2 = 0;
+  int(*local_odd)(int) = local_unsafe;
+  ptr<int(int)> local_super_unsafe1 = local_unsafe; // expected-error {{cannot cast an unchecked function pointer to a checked pointer unless unchecked function is named}}
+  ptr<int(int)> local_super_unsafe2 = (ptr<int(int)>)local_unsafe; // expected-error {{cannot cast an unchecked function pointer to a checked pointer unless unchecked function is named}}
+
 }
--- a/tests/typechecking/function_casts.c
+++ b/tests/typechecking/function_casts.c
@@ -4,6 +4,7 @@
 // RUN: %clang_cc1 -fcheckedc-extension -verify %s
 //
 
+#include <stdint.h>
 #include "../../include/stdchecked.h"
 
 int f0(int a) {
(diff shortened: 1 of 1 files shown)


tests/typechecking/function_casts.c
```diff
@@ -0,0 +1,116 @@
...
+
+  // Unchecked parameter functions are not allowed
+  takes_safe(f);    // expected-error {{cannot cast an unchecked function pointer to a checked pointer unless unchecked function is named}}
+  takes_safe(*f);   // expected-error {{cannot cast an unchecked function pointer to a checked pointer unless unchecked function is named}}
+  takes_safe(&*f);  // expected-error {{cannot cast an unchecked function pointer to a checked pointer unless unchecked function is named}}
+  takes_safe(*&f);  // expected-error {{cannot cast an unchecked function pointer to a checked pointer unless unchecked function is named}}
+  takes_safe(***f); // expected-error {{cannot cast an unchecked function pointer to a checked pointer unless unchecked function is named}}
+
+  // Arbitrary data is definitely not allowed
+  takes_safe((int(*)(int))0xdeadbeef); // expected-error {{cannot cast an unchecked function pointer to a checked pointer unless unchecked function is named}}
```
dtarditi: Please add a test that explicitly casts a function to the wrong function pointer type, not just an integer constant, and then tries to assign that to a ptr function type.  For example:
int g(void);
ptr<int (int)> = (int (*)(int))(g);

Also, please add a test that explicitly casts a value with a checked pointer type to an (incorrect) unchecked function pointer type.

lenary: I believe I have done both. Gladly, they both show the behavior we hope to have.

lenary:
This, I believe, is also ready for another review.

tests/typechecking/function_casts.c
```diff
@@ -0,0 +1,184 @@
...
+ptr<int(int)> allowed_convert1 = f0;
+// Pointers to named functions are specifically allowed
+ptr<int(int)> allowed_convert2 = &f0;
+ptr<int(int)> allowed_convert3 = *f0;
+ptr<int(int)> allowed_convert4 = &*f0;
+ptr<int(int)> allowed_convert5 = *&f0;
+ptr<int(int)> allowed_convert6 = ***f0;
+
+// Arbitrary Data is definitely not allowed
+ptr<int(int)> bad_convert1 = (int(*)(int))0xdeadbeef; // expected-error {{cast to checked function pointer type '_Ptr<int (int)>' must preserve pointer value}}
```
dtarditi: I think this error message could be confusing.

tests/typechecking/function_casts.c
```diff
@@ -0,0 +1,184 @@
...
+  // Top-level declared functions are allowed
+  ptr<int(int)> safe_f1 = f0;
+  ptr<int(int)> safe_f2 = &f0;
+  ptr<int(int)> safe_f3 = *f0;
+  ptr<int(int)> safe_f4 = &*f0;
+  ptr<int(int)> safe_f5 = *&f0;
+  ptr<int(int)> safe_f6 = ***f0;
+
+  // Parameter functions are not allowed
+  ptr<int(int)> unsafe_f1 = f;    // expected-error {{cast to checked function pointer type '_Ptr<int (int)>' may not read memory through unchecked pointers}}
```
dtarditi: This error message is also confusing.

tests/typechecking/function_casts.c
```diff
@@ -0,0 +1,184 @@
...
+ptr<int(int)> allowed_convert4 = &*f0;
+ptr<int(int)> allowed_convert5 = *&f0;
+ptr<int(int)> allowed_convert6 = ***f0;
+
+// Arbitrary Data is definitely not allowed
+ptr<int(int)> bad_convert1 = (int(*)(int))0xdeadbeef; // expected-error {{cast to checked function pointer type '_Ptr<int (int)>' must preserve pointer value}}
+ptr<int(int)> bad_convert2 = (int(*)(int))g0;         // expected-error {{cast to checked function pointer type '_Ptr<int (int)>' from incompatible type 'float (float)'}}
+
+// Now locally within a function body
+void local_convert(int(*f)(int)) {
```
dtarditi: I think it would be helpful to test the positive (allowed) case where f is a checked pointer to a function.

tests/typechecking/function_casts.c
dtarditi: How about "can only cast function names to checked function pointer type"?

lenary:
I think I have addressed all your current concerns.

Issue #101, opened 2017-01-26
Update Syntactic Restrictions on Commas in Non-Modifying Expressions


lenary:
Non-modifying expressions currently disallow all comma expressions. This should be changed to only disallow them at the top level, so they cannot be confused with a function parameter list.

PR #108, opened 2017-02-06
Disallow commas at the top-level of non-modifying expressions.

lenary:
As discussed in person, and in #101.

There's no explicit formal grammar for non-modifying expressions, should there be?


dtarditi:
There's no need to write out a formal grammar for non-modifying expression.   I think saying that they are subset of C expressions and describing the restrictions is clear enough.  It would also likely be an implementation burden to require that they be parsed separately from normal expressions.

PR #109, opened 2017-02-06
Clarifies Function Pointer Casts

lenary:
This brings the specification closer to our compiler's rules for function pointer casts. 

This one will probably need work, specifically around "change the value of" in the language I have added.

### DIFF
--- a/spec/bounds_safety/interoperation.tex
+++ b/spec/bounds_safety/interoperation.tex
@@ -160,12 +160,19 @@ is an incomplete type.
 
 \item Otherwise, if \var{D} is a function pointer type,
 \begin{itemize}
-\item Function names with unchecked function pointer types can be cast to have corresponding checked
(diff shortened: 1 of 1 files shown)


spec/bounds_safety/interoperation.tex
```diff
@@ -162,7 +162,10 @@ \subsection{Description of cast operators}
 \begin{itemize}
 \item Function names with unchecked function pointer types can be cast to have corresponding checked
 pointer types.  If \var{e} is a function name, \var{D} is \ptrT, \var{S} is \uncheckedptrinst{U},
```
dtarditi: I recommend splitting this into a case analysis for greater precision:
Case 1: e is a function name..
Case 2: e has the form * e1 or & e1 where the */& operand does not change the value of e1, and checking succeeds for (D) e1.
Case 3: e is a cast expression.

lenary:
This should be closer to what we're going for, I believe.

PR #110, opened 2017-02-07
Fixes Tests that rely on int->ptr conversions

lenary:
These tests all fail on x86_64.

I think I'm happy enough with coverage that we can get rid of them, but I'll defer to david.

### DIFF
--- a/tests/static_checking/nme_bounds.c
+++ b/tests/static_checking/nme_bounds.c
@@ -216,6 +216,8 @@ void f2(int i, int* loc) {
 
   array_ptr<int> as19a : count(i == j) = 0;
   array_ptr<int> as19b : byte_count(i == j) = 0;
+  array_ptr<int> as19c : bounds(loc + (loc == loc2), loc) = 0;
+  array_ptr<int> as19d : bounds(loc, loc + (loc == loc2)) = 0;
 
   array_ptr<int> as20a : count(i & j) = 0;
   array_ptr<int> as20b : byte_count(i & j) = 0;
--- a/tests/static_checking/nme_bounds.c
+++ b/tests/static_checking/nme_bounds.c
@@ -216,8 +216,6 @@ void f2(int i, int* loc) {
 
   array_ptr<int> as19a : count(i == j) = 0;
   array_ptr<int> as19b : byte_count(i == j) = 0;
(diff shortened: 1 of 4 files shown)


tests/typechecking/function_casts.c
```diff
@@ -193,9 +193,6 @@ void local_convert(int(*f1)(int), ptr<int(int)> f2) {
   ptr<int(int)> local_weird_unsafe1 = (ptr<int(int)>)~(intptr_t)f1; // expected-error {{can only cast function names or null pointers to checked function pointer type '_Ptr<int (int)>'}}
   ptr<int(int)> local_weird_unsafe2 = (ptr<int(int)>)~(intptr_t)f2; // expected-error {{can only cast function names or null pointers to checked function pointer type '_Ptr<int (int)>'}}
   ptr<int(int)> local_weird_unsafe3 = (ptr<int(int)>)~(intptr_t)f0; // expected-error {{can only cast function names or null pointers to checked function pointer type '_Ptr<int (int)>'}}
-  ptr<int(int)> local_weird_unsafe4 = (ptr<int(int)>)!(intptr_t)f1; // expected-error {{can only cast function names or null pointers to checked function pointer type '_Ptr<int (int)>'}}
```
dtarditi: What error message is clang emitting for this case?  I am wondering if all the test cases that involve integer to function pointer casts need to be moved to the platform-specific test subdirectories.

tests/static_checking/nme_bounds.c
```diff
@@ -216,8 +216,6 @@ void f2(int i, int* loc) {
 
   array_ptr<int> as19a : count(i == j) = 0;
   array_ptr<int> as19b : byte_count(i == j) = 0;
-  array_ptr<int> as19c : bounds((int*)(loc == loc2), loc) = 0;
```
lenary: The problem here, as before, is that promoting from `int` (the type of the comparison) to `int*` is a widening operator on some platforms (x86_64 windows at least), and so this triggers a warning. I can put an `// expected-warning` on the line with the message if you'd prefer that?

tests/typechecking/function_casts.c
lenary: This is another `int` to pointer cast, the problem being that pointers on x86_64 are wider than `int`. The testcase just issues a warning. Specifically, the output of the `!` unary operator is always of int type it seems.

PR #119, opened 2017-02-22
Adds the test code for parsing the relative expressoin. (#037)

jijoongmoon:
Adds the test code for relative expression. ( rel_align.c )
 . extract the range bounds expression from declaration_bounds.c, member_bounds.c and return_bounds.c and add rel_align(char).

### DIFF
--- a/tests/parsing/rel_align.c
+++ b/tests/parsing/rel_align.c
@@ -59,8 +59,7 @@ extern void f4(array_ptr<int> arr : count(len), int len) {
   int none = 0;
   array_ptr<int> s : bounds(arr + none, arr + len) rel_align(char) = arr;
   array_ptr<int> t : bounds(t, t + count) rel_align(char) = arr;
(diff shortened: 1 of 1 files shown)


tests/parsing/rel_align.c
```diff
@@ -0,0 +1,219 @@
...
+struct S13 {
+  int len;
+  array_ptr<int> arr : bounds() rel_align(char); //expected-error {{expected expression}} expected-error {{expected ')'}}  expected-note {{to match this '('}}
+};
+
+array_ptr<int> global_bound;
+
+struct S14 {
+  int len;
+  array_ptr<int> arr : bounds(global_bound, global_bound + len) rel_align(char); // expected-error 2 {{use of undeclared member 'global_bound'}} expected-error {{expected ')'}}  expected-note {{to match this '('}}
```
dtarditi: The eorr message about "expected ')'" is spurious.  We should avoid emitting it.

jijoongmoon:
I updated relative alignment bounds clause test examples. I think it is ready for further review.
Thanks.

PR #120, opened 2017-02-24
Tests of Dynamic Checks at Array Subscript and Deref

lenary:
These are tests to go with Microsoft/checkedc-clang#163

### DIFF
--- a/tests/dynamic_checking/deref-arith-check-opt.c
+++ b/tests/dynamic_checking/deref-arith-check-opt.c
@@ -2,6 +2,12 @@
 //
 // The following lines are for the clang automated test suite
 //
+// This builds the file `deref-arith-check.c` at optimisation level O3, outputting an
+// executable at %t. We then run this with the same arguments as we run the unoptimised
+// binary with, and use FileCheck to ensure the output is what we expected, as specified
+// in `deref-arith-check.c`. This is run as a seperate test so we know if optimisation is
+// breaking our dynamic checks.
+//
 // RUN: %clang -fcheckedc-extension %S/deref-arith-check.c -o %t -Werror -O3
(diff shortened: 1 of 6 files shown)


lenary:
I am yet to write tests for deref checks, will do that this afternoon, expect a second commit.

lenary:
I've added the tests for local arrays, will get to parameters soon this afternoon. Also I found I could split the -O3 tests into a separate file (which will run the same code), which keeps the test suite quick.

tests/dynamic_checking/deref-arith-check-opt.c
```diff
@@ -0,0 +1,34 @@
+// Tests that the Dereference Operator works with dynamic checks and pointer arithmetic
+//
+// The following lines are for the clang automated test suite
```
dtarditi: Could you add a comment describing what this file does?   It is important to have a comment explaining that the run lines compile another file other than this one.

tests/dynamic_checking/deref-check-opt.c
```diff
@@ -0,0 +1,17 @@
+// Tests that the Dereference Operator works with dynamic checks
+//
+// The following lines are for the clang automated test suite
```
dtarditi: This too needs a comment describing what this is doing.

tests/dynamic_checking/subscript-check-opt.c
```diff
@@ -0,0 +1,37 @@
+// Tests that the Array Subscript Operator works with dynamic checks
+//
+// In particular, this re-runs the tests in subscript-check.c with Optimizer at O3
```
dtarditi: This needs to point out that that the RUN line actually compilers another file other than this one.

PR #121, opened 2017-03-02
Fixes issues with checked system headers on Mac with ifdefs

lenary:
This is a first pass at making the build pass on mac.

Apple defines the `__APPLE__` constant, which we can use to find out if we're on mac. For the moment I'm just not defining our checked versions of these stdlib functions if we're on a mac, I want to circle back around to getting the definitions correct, but this may take some time because of how Apple have structured their C system headers. 

Expect another commit which provides alternative definitions which we can actually use on Macs.

### DIFF
--- a/include/_builtin_stdio_checked.h
+++ b/include/_builtin_stdio_checked.h
@@ -1,0 +1,57 @@
+//---------------------------------------------------------------------//
+// Bounds-safe interfaces for compiler-defined builtin functions       //
+// corresponding to stdio.h functions                                  //
+//                                                                     //
+// These are given in the order they appear in clang's Builtins.def.   //
+// Functions that do not appear can not have checked interfaces        //
+// defined.                                                            //
+//                                                                     //
+// These are based on the types as declared within clang               //
+// and https://gcc.gnu.org/onlinedocs/gcc/Object-Size-Checking.html    //
+//                                                                     //
+// TODO: revise string types after support for pointers to             //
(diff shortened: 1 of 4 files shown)


lenary:
So, there won't be a follow-up commit. We run into a problem where in some cases apple swaps out the underlying definition into a macro invocation which invokes `__builtin_<function>_chk` for some functions.

This can be seen here: https://opensource.apple.com/source/Libc/Libc-1158.30.7/include/secure/_string.h.auto.html

Here is a page documenting what these builtin functions do, for GCC. Clang's equivalents are undocumented. https://gcc.gnu.org/onlinedocs/gcc/Object-Size-Checking.html

lenary:
Ok, actually, there is a simpler fix, which is to `#define _FORTIFY_SOURCE 0` at the top of any of these files. But that disables any compiler bounds checks that clang might insert in `__builtin_<fn>_chk`.

The advantage of this latter fix is that we don't tangle the headers with a lot of ifdefs.

lenary:
I believe this is a reasonable fix, I'd like you to check my logic is sound (I'm also going to run these on linux). 

The string versions just add another size_t argument for the object size. 

The printf-like versions add two arguments, an int flag, and a size_t argument for the object size. The printf-like versions also have a function that hangs around without the `builtin` prefix, which we can add bounds to. This is where specific handling of the `flag` argument happens.

PR #123, opened 2017-03-07
Use signals for out-of-bound memory access tests.

dtarditi:
This way processes exit gracefully, instead of relying on the OS to handle
a crash.  On Windows, crashes cause Windows Error Reporting (WER) to fire up,
invoking system logging.

We have been using the not program from the LLVM test harness with the --crash
option to determine when processes have exited with a crash.  We have found
that this approach is not working reliably on Windows.  The not program spawns
a child process and looks at the process exit code, looking for negative exit
codes to indicate a crash. When running automated testing under Visual Studio
Team Services, sometimes the exit code is a positive number.

Testing:
- Tested this change on Windows and Linux and all the Checked C tests passed as expected.

### DIFF
--- a/tests/dynamic_checking/deref-check.c
+++ b/tests/dynamic_checking/deref-check.c
@@ -5,15 +5,16 @@
 // RUN: %clang -fcheckedc-extension %s -o %t -Werror
 // RUN: %t pass1 | FileCheck %s --check-prefixes=CHECK,CHECK-PASS,CHECK-PASS-1
 // RUN: %t pass2 | FileCheck %s --check-prefixes=CHECK,CHECK-PASS,CHECK-PASS-2
-// RUN: not --crash %t fail1 | FileCheck %s --check-prefixes=CHECK,CHECK-FAIL,CHECK-FAIL-1
-// RUN: not --crash %t fail2 | FileCheck %s --check-prefixes=CHECK,CHECK-FAIL,CHECK-FAIL-2
-// RUN: not --crash %t fail3 | FileCheck %s --check-prefixes=CHECK,CHECK-FAIL,CHECK-FAIL-3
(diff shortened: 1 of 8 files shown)


tests/dynamic_checking/deref-check.c
```diff
@@ -5,15 +5,16 @@
...
-// RUN: not --crash %t fail1 | FileCheck %s --check-prefixes=CHECK,CHECK-FAIL,CHECK-FAIL-1
-// RUN: not --crash %t fail2 | FileCheck %s --check-prefixes=CHECK,CHECK-FAIL,CHECK-FAIL-2
-// RUN: not --crash %t fail3 | FileCheck %s --check-prefixes=CHECK,CHECK-FAIL,CHECK-FAIL-3
-// RUN: not --crash %t fail4 | FileCheck %s --check-prefixes=CHECK,CHECK-FAIL,CHECK-FAIL-4
+// RUN: %t fail1 | FileCheck %s --check-prefixes=CHECK,CHECK-FAIL,CHECK-FAIL-1
+// RUN: %t fail2 | FileCheck %s --check-prefixes=CHECK,CHECK-FAIL,CHECK-FAIL-2
+// RUN: %t fail3 | FileCheck %s --check-prefixes=CHECK,CHECK-FAIL,CHECK-FAIL-3
+// RUN: %t fail4 | FileCheck %s --check-prefixes=CHECK,CHECK-FAIL,CHECK-FAIL-4
 
+#include <assert.h>
```
lenary: I don't think you meant to include this header.

dtarditi: It was already there - I just alphabetized the includes of standard header files.

lenary: Oh so it was. I'm helping. LGTM, merge when ready.

PR #126, opened 2017-03-09
Dynamic bounds check array update

lenary:
This adds high-level tests of compound assignment and increment operators to our codebase.

Given these tests are all very similar, we saw fit to document the structure of the tests in this directory for future maintainers/testers.

### DIFF
--- a/tests/dynamic_checking/increment-deref-check.c
+++ b/tests/dynamic_checking/increment-deref-check.c
@@ -140,10 +140,11 @@ void passing_test_3(array_ptr<int> a : count(len), int len) {
 
 // Bounds describe empty range, no deref
 void failing_test_1(void) {
-  array_ptr<int> a : bounds(a, a) = 0;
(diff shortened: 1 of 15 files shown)


tests/dynamic_checking/increment-deref-check.c
```diff
@@ -0,0 +1,178 @@
...
+
+  (*a)++;
+  printf("Assignable: %d\n", *a);
+
+  puts("Expected Success");
+}
+
+// Bounds describe empty range, no deref
+void failing_test_1(void) {
+  array_ptr<int> a : bounds(a, a) = 0;
```
dtarditi: In the Checked C semantics, wouldn't this fail first because a is null?   We haven't implemented null checking yet, but when we do, I don't think this test will be checking what the comment says it is checking.

lenary: Ah, yes, I should go back and make sure these all do the right thing in normal C.

Issue #127, opened 2017-03-17
Remove error-check in rel_align.c


jijoongmoon:
Remove error-check in rel_align.c to along with "Avoiding generating error message 'expected bounds expression' when there is a typechecking error (checkedc-clang : #195)

PR #128, opened 2017-03-17
Remove error-check in rel_align.c to get along with "Avoiding generat…

jijoongmoon:
…ing error message 'expected bounds expression' when there is a typechecking error [ checkedc-clang : #195 ]" (#127)


jijoongmoon:
I think it is ready to review. Thanks always!

PR #129, opened 2017-03-17
+ Add test code for checked scope implementation (#189)

wonsubkim:
- simple version of checked scope parsing & type checking
  - checked function
   - checked keyword is before function definition and it is used as function specifier
   - in checked function, it checks if function declaration(return & parameter) is checked scope
   - in checked function, it checks if function definition(function statement boyd) is checked scope
   - in checked function, function with no prototype is not allowed
  - checked scope
   - checked keyword is before l_brace and corresponding compound statments are affected by checked scope
   - in checked scope, checked pointer or unchecked pointer with bounds-safe interface is only allowed
   - in checked scope, child scope inherits parent checked property

### DIFF
--- a/tests/typechecking/checked_scope.c
+++ b/tests/typechecking/checked_scope.c
@@ -23,7 +23,7 @@ checked int* f3() : itype(array_ptr<int>) { //expected-error {{return cannot hav
 }
 checked int* f4(int len, int *p, int *q, array_ptr<int> r) : itype(ptr<int>) { // expected-error 2 {{parameter cannot have an unchecked pointer type}}
 }
-checked int f5(int p[]) { // expected-error {{parameter cannot have an unchecked pointer type in a checked scope}}
+checked int f5(int p[]) { // expected-error {{parameter cannot have an unchecked array type in a checked scope}}
   int a = 5;
   int *upa = &a; // expected-error {{variable cannot have an unchecked pointer type in a checked scope}}
   ptr<int> pb = &a;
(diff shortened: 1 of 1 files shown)


wonsubkim:
+ Not implemented test
 - structure member variable
 - C-style unchecked pointer type casting
 - change type produced by address of operator

tests/typechecking/checked_scope.c
```diff
@@ -0,0 +1,240 @@
...
+  int *upa; // expected-error {{variable cannot have an unchecked pointer type in a checked scope}}
+  ptr<int> pb = 0;
+  array_ptr<int> pc;
+  return upa;
+}
+
+checked int* f16(int a checked[][5], int b checked[][5]) : itype(ptr<int>) checked {
+}
+
+checked int* f17(int a [][5], int b [][5]) : itype(ptr<int>) checked { // expected-error 2 {{parameter cannot have an unchecked pointer type in a checked scope}}
```
dtarditi: I think we want to change this error message for array parameters.   Clang has inserted a conversion from array type to pointer type and retyped each parameter variable as a pointer type.   ParmVarDecl has a method called getOriginalType() that will get the original declared type for a variable.   You can use it in `Sema::DiagnoseCheckedDecl`. It might make the error message easier for programmers to understand.

wonsubkim: Thanks for this suggestion. I did not realize it
I will apply this change after completing test

PR #130, opened 2017-03-21
Add test code to check the issue #188 "Improve error recovery for bou…

jijoongmoon:
…nds expression parsing".

 ex> array_ptr<int> arg : bo0unds(arg, arg+5) rel_align(1) = 0;
   . bo0unds : expect bounds expression error
   . rel_align : expect a type error

### DIFF
--- a/tests/parsing/rel_align.c
+++ b/tests/parsing/rel_align.c
@@ -193,14 +193,14 @@ extern void f16(void) {
       : bounds(arr, arr + len) rel_align(char);
 }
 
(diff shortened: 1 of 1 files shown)


jijoongmoon:
I think it is ready for the review. Thanks always.

jijoongmoon:
Oh. yes. I ran it with "make check-checkedc". and it passed except samples.
I think if there is right relative bounds clause after misspelled bounds, it shows only "expected bounds expression". If there is error such as "rel_align(1)" after misspelled bounds, then it shows the error "expected a type" and "expected bounds expression" error.

PR #131, opened 2017-03-28
Add & modify test code for checked scope feature (#189) and related work-item(#201, #202)

wonsubkim:
Test code is modified to apply checkedc-clang modification for checked block implementation

1. implicit cast rule has changed to allow implicit cast between checked pointer types
Therefore, previous expected-error is changed to expected-correct
pointer_types.c & checked_arrays.c is modified file in checkedc test program

2. related checked_scope test code is extended
 - check if No-use of unchecked pointer/array type in checked block
 - check if No-use of unchecked pointer type casting in checked block
 - changed semantic of address-of operator in checked block

### DIFF
--- a/tests/typechecking/checked_arrays.c
+++ b/tests/typechecking/checked_arrays.c
@@ -289,9 +289,9 @@ extern void check_assign_cv(int param[10],
   p = checked_param;              // expected-error {{incompatible type}}
   p_const = checked_param;        // expected-error {{incompatible type}}
   p_volatile = checked_param;     // expected-error {{incompatible type}} 
-  q = checked_param;              // expected-correct {{incompatible type}}, ptr<T> = array_ptr<T> OK
-  q_const = checked_param;        // expected-correct {{incompatible type}}, ptr<T> = array_ptr<T> OK
-  q_volatile = checked_param;     // expected-correct {{incompatible type}}, ptr<T> = array_ptr<T> OK
+  q = checked_param;              // ptr<T> = array_ptr<T> OK
+  q_const = checked_param;        // ptr<T> = array_ptr<T> OK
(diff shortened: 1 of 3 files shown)


tests/typechecking/checked_arrays.c
```diff
@@ -289,9 +289,9 @@ extern void check_assign_cv(int param[10],
   p = checked_param;              // expected-error {{incompatible type}}
   p_const = checked_param;        // expected-error {{incompatible type}}
   p_volatile = checked_param;     // expected-error {{incompatible type}} 
-  q = checked_param;              // expected-error {{incompatible type}}
-  q_const = checked_param;        // expected-error {{incompatible type}}
-  q_volatile = checked_param;     // expected-error {{incompatible type}}
+  q = checked_param;              // expected-correct {{incompatible type}}, ptr<T> = array_ptr<T> OK
```
lenary: In this case, change the comment to `// ptr<T> = array_ptr<T> OK`. `expected-<kind> {{<message>}}` in comments is being searched for by the diagnostic verifier (enabled by the argument `-verify` above). As this behavior is allowed, there should be no diagnostics (warnings/errors/etc). Later on where you've added `expected-error {{...}}`, the comment should just be the `expected-error {{...}}` part of the comment.

wonsubkim: Okay. I agree your feedback
If the behavior is allowed, I will remove related comment.

wonsubkim:
I will modify error messages and some strategies for checking error according to your feedback
After changing check-clang code, I will also change testing codes
Thanks

dtarditi:
I will wait to merge this until after the corresponding compiler pull request has been approved.

Issue #135, opened 2017-04-03
Simplify and extend test code in checkedc_scope_pragma.c


dtarditi:
In pull request #203, I wrote:

> 2100 lines of test code is a lot code to test the BOUNDS_CHECKED pragma.    This may be costly to maintain in the future.   Please simplify this test code to be more concise.  I recommend selective checking properties that you'd expect to be true with a checked or unchecked scope.    

> For example, you already test that unchecked parameters are not allowed when BOUNDS_CHECKED has been turned on.   It would be good to have a few tests of that.  I don't think we need as many as you have.  You also check that unchecked local variables are not allowed when BOUNDS_CHECKED has been turned on.  It would be good to also have a few tests of that.

> Pragmas can also be turned on and off and I believe can be limited to specific scopes.    Could you check that too?

So that I can commit changes to the compiler,  I'm merging the pull request.  Please revise the code for the BOUNDS_CHECKED pragma.

PR #136, opened 2017-04-04
simplify and extend test code in checked_scope_pragma.c (#135)

wonsubkim:
+ it works well with corresponding latest checkedc-clang implementation
  + simplify test code for BOUNDS_CHECKED pragma
  + #pragma BOUNDS_CHECKED on
  + In checked scope of compilation unit
  Test unchecked function specifier
  By default, all functions are checked function
  By default, all global variables are checked type
  By default, all structures are checked type
  Function parameter & return values SHOULD be checked type in checked scope of compilation unit
  Also, local variables, global variables, structure members SHOULD be checked type

+ modify specification related to BOUNDS_CHECKED pragma value
  + BOUNDS_CHECKED on-off-switch value can be [on|off]
  + default value is off

### DIFF
--- a/spec/bounds_safety/core-extensions.tex
+++ b/spec/bounds_safety/core-extensions.tex
@@ -740,7 +740,8 @@ the top-level scope is a checked scope:
 #pragma BOUNDS_CHECKED \textit{on-off-switch}
 \end{alltt}
 
-Where \texttt{\textit{on-off-switch}} is one of \verb|ON OFF DEFAULT|
+Where \texttt{\textit{on-off-switch}} is one of \verb|on off|
+and it is \verb|off| by default
 
 By default, function definitions are not checked. A block inherits the
 checking properties of its parent. This preserves the meaning of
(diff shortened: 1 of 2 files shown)


wonsubkim:
If you turned off the pragma, you can see that it has different kinds of errors

If pragma is turned on, it expects type A expected-error message
Otherwise, it expects another kind of error message
Do you have any idea to generate diagonostic error messages for different BOUNDS_CHECKED value?

dtarditi:
When using the LLVM lit infrastructure, you can specify different prefixes for different test runs that produce different errors.  Some of Sam Elliott's tests do this.  They are in clang\tests\CheckedC\dynamic-checks.  So  you can have something like:

CHECKON: .... expected error message
CHECKOFF: ... expected error message.

You specify the prefix as part of your FileCheck command using the `--checked-prefix' command.  You specifiy something like '--check-prefix=CHECKOFF`.

lenary:
The documentation for the `--check-prefix` argument is here, it's an option on the FileCheck executable: http://www.llvm.org/docs/CommandGuide/FileCheck.html

PR #138, opened 2017-04-05
Add test code for dynamic and assume bounds cast (#204)

jijoongmoon:
This is the test code for the pointer bounds casting. (#204)
- It will be added more test cases.


jijoongmoon:
I think it is needed to test the pointer bounds cast in checkedc-clang.

jijoongmoon:
I add more test code for Pointer Bounds Cast which includes assume_bounds_cast<ptr<int>> and assume_bounds<int *>. It's ready for further review. Thanks always.

PR #139, opened 2017-04-05
 Add test code for variable argument function call issue in checked block (#251) & BOUNDS_CHECKED pragma (#247)

wonsubkim:
+  Add test code for variable argument function call in checked block (#251)
  + Variable argument function call is not allowed in checked blocks
  + variadic function call & printf function call is checked
+ Add test code for BOUNDS_CHECKED pragma (#247)
  + For elaborate test BOUNDS_CHECKED pragma with a scope, add test cases
  + For elaborate scope handling, implementation is changed to use annotation token
  By using annotation token, it can recognize scope corresponding to annotation token
+ Restore on-off-switch syntax in pragma BOUNDS_CHECKED
  + #pragma BOUNDS_CHECKED [on-off-switch]
  + on-off-switch = ON/OFF/DEFAULT

### DIFF
--- a/spec/bounds_safety/core-extensions.tex
+++ b/spec/bounds_safety/core-extensions.tex
@@ -740,8 +740,7 @@ the top-level scope is a checked scope:
 #pragma BOUNDS_CHECKED \textit{on-off-switch}
 \end{alltt}
 
-Where \texttt{\textit{on-off-switch}} is one of \verb|on off|
-and it is \verb|off| by default
+Where \texttt{\textit{on-off-switch}} is one of \verb|ON OFF DEFAULT|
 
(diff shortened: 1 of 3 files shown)


wonsubkim:
I have added test code for issue ( #142 )
Constructed type that contains unchecked pointer is not allowed since it can mislead pointer usage.
For incorrect constructed type usage, I have added expected-error
`ptr<int [10]>` - checked pointer to unchecked pointer is treated as error case

variable arguments type is also processed recursively in same manner to handle constructed type
`ptr<int(int, ...)>` - function pointer type variable having variable arguments is treated as error case.

wonsubkim:
I am ready for review :)

PR #144, opened 2017-04-10
Remove span type from the language extension design.

dtarditi:
The span type seems like a type that is better suited for a library than a core
language extension. In C++, which has templates, there is already a proposal
for adding a span-like type to the C++ Standard Library.  Remove the span
type for the language extension design and move it to the design alternatives
section.

To be clear, we are postponing adding the span type to the language extension
design (and the clang implementation as well).   We may decide to add it in
the future, especially because implicit conversions involving library types
and pointers can be messy.  However, for now,  it can wait until the need
becomes clearer.

This commit also changes the version description to "Version 0.7 - Draft" and
changes the returned by the address-of operator in checked block from ptr
to array_ptr.

### DIFF
--- a/spec/bounds_safety/introduction.tex
+++ b/spec/bounds_safety/introduction.tex
@@ -472,7 +472,8 @@ and Chapters~\ref{chapter:related-work} and \ref{chapter:design-alternatives}.
 
 This design has benefited from many discussions with Weidong Cui, Gabriel Dos Reis, 
 Chris Hawblitzel, Galen Hunt, Shuvendu Lahiri, and Reuben Olinsky.  The design has
-benefited also from feedback from Michael Hicks, Wonsub Kim, Greg Morrisett, Jonghyun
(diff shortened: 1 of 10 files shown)


spec/bounds_safety/introduction.tex
```diff
@@ -473,8 +472,7 @@ \section{Acknowledgements}
 
 This design has benefited from many discussions with Weidong Cui, Gabriel Dos Reis, 
 Chris Hawblitzel, Galen Hunt, Shuvendu Lahiri, and Reuben Olinsky.  The design has
-benefited also from feedback from Sam Elliott, Michael Hicks, Wonsub Kim, Greg
-Morrisett, Jonghyun Park, and Andrew Ruef. We thank them for their contributions to
-the design.
+benefited also from feedback from Michael Hicks, Wonsub Kim, Greg Morrisett, Jonghyun
```
lenary: This hunk is unrelated. Please don't erase me :(

PR #146, opened 2017-04-11
Add test code for relative alignment clause in Bounds Cast Expression…

jijoongmoon:
…. (#262)


jijoongmoon:
I think it is ready to review. I will add more test codes shortly. Thanks always.

PR #151, opened 2017-04-25
Modify error check in rel_align.c to check #282 issue.

jijoongmoon:
: Fix Return value of ParseRelativeBoundsForDecl not being used (#282)

### DIFF
--- a/tests/parsing/rel_align.c
+++ b/tests/parsing/rel_align.c
@@ -193,7 +193,7 @@ extern void f16(void) {
       : bounds(arr, arr + len) rel_align(char);
 }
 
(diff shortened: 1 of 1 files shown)


jijoongmoon:
I think it is ready for review. Thanks.

jijoongmoon:
I think it is ready for further review. Thanks always!

dtarditi:
I don't think there are actually any test changes here - these look like white space changes.   I think we can abandon this pull request. 

I thought a bit about how to test the corresponding compiler change and I can't figure out an easy way right now.    The corresponding compiler change suppresses downstream errors when the relative alignment clause is invalid.  We would end up with a bounds declaration that looked valid, but actually had no relative alignment clause when it was supposed to have one.  That could trigger errors during checking of bounds declarations.  We haven't implemented checking of bounds declarations yet, so there is nothing to check right now.

jijoongmoon:
Yes. You are right. I will close this pull request.

PR #153, opened 2017-04-28
Add test code for (checkedc-clang/#277) issue

wonsubkim:
+ Add test code for (checkedc-clang/#277) issue
  + prevents bounds cast operation to unchecked type in a checked scope
  + prevents _Assume_bounds_cast operation in a checked scope

### DIFF
--- a/tests/typechecking/checked_scope.c
+++ b/tests/typechecking/checked_scope.c
@@ -1607,17 +1607,25 @@ checked void check_cast_operator(void) {
   ptr<int> q = 0;
   array_ptr<int> r : count(5) = 0;
   checked {
-  /* DestType : unchecked, Src->Type : ptr */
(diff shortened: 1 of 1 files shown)


tests/typechecking/checked_scope.c
```diff
@@ -1601,6 +1601,25 @@ checked void check_cast_operator(void) {
   gptr2 = upa;                      // expected-error {{cannot use a variable with an unchecked type}}
   gptr2 = (ptr<int>)upa;            // expected-error {{cannot use a variable with an unchecked type}}
   gptr2 = (array_ptr<int>)upa;      // expected-error {{cannot use a variable with an unchecked type}}
+
+  unchecked {
+  int *p;
+  ptr<int> q = 0;
+  array_ptr<int> r : count(5) = 0;
+  checked {
+  /* DestType : unchecked, Src->Type : ptr */
```
dtarditi: There needs to be a test for the case where the target type of the cast is a variable argument function type.

PR #156, opened 2017-05-02
Add test code for bounds cast expression (#286)

jijoongmoon:
. Add syntax check for *_bounds cast expression with bounds expressions.
   ex)
   array_ptr<int> x : count(10)
       = _Dynamic_bounds_cast<array_ptr<int>>(x, count(10));
   array_ptr<int> x : count(10)
       = _Dynamic_bounds_cast<array_ptr<int>>(x, bounds(a, b));
 . Add test code to check whether E1 is pointer type or integral type.


jijoongmoon:
I think it is ready for review. Thanks!

dtarditi:
I will wait to merge this until the compiler change are ready.

Issue #154, opened 2017-04-28
Add runtime tests of bounds checking for struct operations.


dtarditi:
Structure members may have bounds or may be checked arrays.    In those cases, bounds checks are needed when the value of the member is used to read to write memory.   

We have runtime tests of bounds checking for some combinations of expressions that produce lvalues that are then used to read or write memory.    We need to add more tests of combinations of expressions involving struct member access operators (`.` and `->`).

PR #157, opened 2017-05-08
Expand and reorganize dynamic bounds checking tests.

dtarditi:
This adds tests of pointer dereference (`*`) and array-subscript operators (`[]`) applied to member reference expressions with pointer type.  This addresses issue #154.  For example:
```
struct S v ...
struct S *pv ...
*(v.f)
*(pv->f)
v.f[5]
pv->f[5]
```
The number of test files started to get out-of-hand, so I merged all the existing bounds checking tests (which had 4 different variants) into single files.  I used pre-processing to cut back on the redundancy.  Now we defined a macro to get a specific variant.  The README.md file is updated to describe this new organization.

The GitHub diff mechanism isn't very helpful with changes of this extent.

Testing:
Revised Checked C tests pass on x86 Windows.


tests/dynamic_checking/bounds/deref.c
```diff
@@ -0,0 +1,247 @@
...
+
+#include <assert.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include "../../../include/stdchecked.h"
+
+#ifdef TEST_READ
+#define TEST_OP(e1, e2)
```
lenary: I don't understand how this works, if this macro expands to nothing. Or do I misunderstand macro definitions without bodies in C?

tests/dynamic_checking/bounds/deref.c
```diff
@@ -0,0 +1,247 @@
...
+  puts("Unexpected Success");
+}
+
+// Bounds describe empty range (a + 2 > a), no deref
+void failing_test_2(void) {
+  int a checked[3] = { 0, 0, 0 };
+  array_ptr<int> b : bounds(a + 2, a) = a;
+
+  TEST_OP(*b, 2);
+  *b = 2;
```
lenary: Why do you need this assign after `TEST_OP(*b, 2)`? Surely the test should fail at TEST_OP if it's going to fail, and the assignment is useless? 

Is it something to do with code not being optimised away?

tests/dynamic_checking/bounds/deref.c
dtarditi: Yes, the macro expands to nothing.  This leads to the elision of any operation.    The tests were counting on the value being read by the subsequent print statement.  That let me avoid having to suppress warnings about unused values.  I agree this is obscure, so I've pushed a change where TEST_READ expands to read the expression.

tests/dynamic_checking/bounds/deref.c
dtarditi: This was an oversight.  I'll fix it.

tests/dynamic_checking/bounds/deref.c
lenary: Ah, now I understand, but I agree that actually having the read happen is a good idea.

dtarditi:
The concern about lack of parallelism is a real concern.  I measured elapsed time for the Checked C regression tests on my 12 core machine Windows machine. This change more than doubled it.  It went from 30 seconds to 106 seconds.

We are doing more testing, so I'd expect the time to increase, but not that much.   For debug versions of clang, clang compiles are definitely the bottleneck for testing.  We had 30 invocations of clang outside of the bounds checking tests and 24 for the bounds checking dynamic tests before, for 54 in all.   We now have 56 invocations of clang for the bounds checking dynamic tests, so total clang invocations goes to 86.  So with a back of the envelope calculations, I'd guess that a 50-60% increase is reasonable,

I want to avoid a proliferation of small test driver files.   The member_expr tests do twice as much work as the other tests files because they test both `.` and `->` operators using RUN commands in file.   I'll see what happens if I create test driver files for one of the operators.

dtarditi:
Thanks for the review!  I'll commit this and address the parallelism issues separately then.

PR #161, opened 2017-05-11
Split tests for bounds checking member expression  into more files.

dtarditi:
This improves the parallelism of running the Checked C regression tests.  The tests for bounds checking of member expressions were doing 8 different compiles: 4 for the dot operator and 4 for the arrow operators.  This change creates separate test automation driver files for the arrow operators.

The elapsed time for a test run on my 12 core Windows machine drops from 68 seconds to 45 seconds.  The original time before new tests were added was 30 seconds.  From back-of-the-envelope calculations, we would expect about a 50-60% increase in testing time with the additional tests.    This changes brings testing time into that range.

Testing:
- Passed automation testing on Linux
- Build and tested locally on Windows.


tests/dynamic_checking/bounds/deref_arrow_member_expr_opt.c
```diff
@@ -0,0 +1,66 @@
...
+// expressions.
+//
+// This builds the file `deref_dot_member_expr.c` at optimisation level O3 
+// with -DARROW_OPERATOR defined, outputting executables at %tn. We then run 
+// these with the same arguments as we run the unoptimised binaries with, and
+// use FileCheck to ensure the output is what we expected, as specified in 
+// `subscript_dot_member_expr.c`. This is run as a separate test so we know
+// if optimisation is breaking some dynamic checks.
+//
+// RUN: %clang -fcheckedc-extension %S/deref_dot_member_expr.c -o %t1 -DTEST_READ -DARROW_OPERATOR -Wer/ror -Wno-unused-value -O3
```
lenary: Typo in `-Werror`

PR #165, opened 2017-06-01
Cast to checked array pointers in bounds declarations

lenary:
This is the follow-on to Microsoft/checkedc-clang#311

This should solve any problems relating to casts to unchecked pointers in checked scopes.

I'll do a second PR where I port this change to the test suite and see if we can remove some more unchecked lines/functions.

### DIFF
--- a/include/_builtin_common.h
+++ b/include/_builtin_common.h
@@ -1,0 +1,29 @@
+//---------------------------------------------------------------------//
+// Bounds-safe interfaces for compiler-defined builtin functions       //
+// corresponding to secure/_common.h functions                         //
+//                                                                     //
+// These are given in the order they appear in clang's Builtins.def.   //
+// Functions that do not appear can not have checked interfaces        //
+// defined.                                                            //
+//                                                                     //
+// These are based on the types as declared within clang               //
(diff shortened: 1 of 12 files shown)


lenary:
Turns out I have more changes which are backports of the header changes we did at the last minute for the deadline. Incoming shortly.

lenary:
This should again be ready for review. This adds `_Unchecked` to any function with a string-y `char*` parameter, and adds `#pragma BOUNDS_CHECKED ON` around every header we can. 

Testing:
Ran the following in the include directory, to check we got no errors or warnings:
`for f in [^_]*.h; do clang -I. -include $f -fcheckedc-extension -c -x c /dev/null -D_CHECKEDC_MOCKUP_THREADS; done`

lenary:
Oh, I've been thinking about these pragmas wrong, hence the bug. Yes, the logic is definitely wrong, and we should do some kind of push/pop, I'll take a look at that later.

It seems [clang already has a system for pushing/popping diagnostic pragmas][1], we might want to echo it, and then yes, add some kind of `ON`/`OFF`/`STRICT` setting (where `STRICT` doesn't let you change it inside that scope).

[1]: https://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-via-pragmas

PR #166, opened 2017-06-01
strncmp is broken on Linux

lenary:
On linux (specifically Ubuntu 14.04), `strncmp` is a macro rather than a declaration once you've included `<string.h>` (it starts as a declaration, then is replaced later on).

This then breaks our redeclaration. So for the moment, let's not redeclare strncmp.

### DIFF
--- a/include/string_checked.h
+++ b/include/string_checked.h
@@ -59,7 +59,13 @@ int memcmp(const void *src1 : byte_count(n), const void *src2 : byte_count(n),
 // int strcmp(const char *src1, const char *src2);
 // int strcoll(const char *src1, const char *src2);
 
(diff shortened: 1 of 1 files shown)


dtarditi:
PR #168 includes an alternate fix for this issue.  We can simply #undef the macro.  This is suggested by the C11 standard in Section 7.1.4. as a solution for the case where someone has created a macro with the same name as a function.

lenary:
Yeah, it seems like any solution is going to be sub-optimal (if the strncmp macro is what makes sure clang can do constant folding), but also `undef`-ing seems more sound than not adding the checked declaration.

We should probably work out a way of undef-ing then re-defing, but we can do that later.

PR #167, opened 2017-06-03
Header Fixups from Benchmark Conversion

lenary:
A Variety of changes

- Removing some `_Unchecked` annotations
- Adding interop type to `__builtin_object_size`
- Adding interop types for `stdin` `stdout` and `stderr`
- Adding interop declaration for `fputc`

There will be more changes incoming, as I complete other benchmark conversions. This branch is what I'm using when testing my changes.


lenary:
I rebased this on top of our changes to the header infrastructure, so it needs re-reviewing.

PR #168, opened 2017-06-05
Copy Checked C headers to clang system header directory.

dtarditi:
This adds a CMake file for copying the Checked C headers into the
clang system header directory as part of the clang build.  This avoids
people having to copy the Checked C header files to their specific project,
which would be error prone and result in synchronization problems.   This
change also fixes an issue with strncmp on Linux by undefining a problematic
macro.

It is a little clunky to store all of these header files in the clang system
header directory because it includes wrapper files that are not needed
by the compiler.  This seems to be better and simpler than the
alternatives, so we're using it.

This updates all the samples in the Checked C repo to no longer
use the full relative path the stdchecked.h.  We can just rely on
it being picked up as part of the system search path.

This change has matching changes for the Checked C LLVM repo, the
Checked C clang repo, and the Checked C LLVM Test Suite Repo.

Testing:
- Passed automation testing for Linux x64 LNT testing  and Windows
x86 when the testing was pointed at the matching branches in all
the repos.

### DIFF
--- a/include/CMakeLists.txt
+++ b/include/CMakeLists.txt
@@ -3,7 +3,7 @@
 # directory for clang so that they can easily be found/used.
 #
 # Checked C does not require runtime system changes, and it would
-# by unnecesarily heavyweight to require people to pick up a new
+# be unnecesarily heavyweight to require people to pick up a new
 # library just to get the Checked C include wrpaper files.
 # So we place the header files with the compiler.
 
@@ -23,7 +23,7 @@ set(files
   )
 
 # Hack - compute the CLANG version from the LLVM version.  The
-# CLANG_VERSION variable is not available at this during the build.
+# CLANG_VERSION variable is not available at this point during the build.
 
 string(REGEX MATCH "[0-9]+\\.[0-9]+(\\.[0-9]+)?" CHECKEDC_CLANG_VERSION
   ${PACKAGE_VERSION})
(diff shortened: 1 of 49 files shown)


include/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+  stdlib_checked.h
+  string_checked.h
+  threads_checked.h
+  time_checked.h
+  _builtin_stdio_checked.h
+  _builtin_string_checked.h
+  )
+
+# Hack - compute the CLANG version from the LLVM version.  The
+# CLANG_VERSION variable is not available at this during the build.
```
lenary: missing word "not available at this **point** during the build."

dtarditi:
I've address the code review feedback.

include/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+  math_checked.h
+  signal_checked.h
+  stdchecked.h
+  stdio_checked.h
+  stdlib_checked.h
+  string_checked.h
+  threads_checked.h
+  time_checked.h
+  _builtin_stdio_checked.h
+  _builtin_string_checked.h
```
dtarditi: Cmake documentation advises against using GLOB: https://cmake.org/cmake/help/v3.9/command/file.html

lenary: That's infuriating, but i guess it means it's explicit when we add new headers.

Issue #169, opened 2017-06-07
Checked C header files cause build error on Windows


dtarditi:
PR #167 add bounds-safe interface types for stdin stdout, and stderr.   On Windows, this is causing a build error.

dtarditi:
This is fixed.

PR #170, opened 2017-06-07
Fix stdio_checked.h for Windows.

dtarditi:
Stdio_checked.h is failing to compile on Windows (issue #169).  Bounds-safe
interfaces were added for stdin, stdout, and stderr.  These identifiers only
have to be expressions with type FILE *.  On Windows, they are function calls,
so we add a bounds-safe interface for the function being called.

### DIFF
--- a/include/stdio_checked.h
+++ b/include/stdio_checked.h
@@ -15,9 +15,16 @@
 
 #pragma BOUNDS_CHECKED ON
 
+#if defined(_WIN32) || defined(_WIN64)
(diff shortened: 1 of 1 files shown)


dtarditi:
I committed this to fix test automation failures.  Sam, feel free to review and suggest changes.

lenary:
Sure, I was running tests, I don't know why I forgot to check the results. 

I think we may need a more principled approach to splitting up platform-specific definitions like this, but that can wait until after I've written this report.

PR #174, opened 2017-06-16
Correct Conditionals for apple system header support

lenary:
I should not be trusted with De Morgan's Laws, turns out I'll break them.

`!a && b` is not equivalent to `!(a && b)`.

This should make anagram work on Linux again, I tested it on my WSL installation


lenary:
I'm going to merge now so I can run a build before I go home.

PR #175, opened 2017-06-19
Test cases for parsing generic function calls

jp1729:
These two files are test cases for parsing generic function calls.
The tests include
* Does clang parse generic function call?
* Does clang correctly detect syntax errors?
* In one instance of testing, I have found that having a previous "error" makes the Expr* that contain the functionDecl to be NULL. This broke my initial implementation, and crashed the compiler. So I fixed the implementation, and added a test to make sure that my new implementation will not crash the compiler.

### DIFF
--- a/tests/parsing/typevariable/generic_func_parsing_error.c
+++ b/tests/parsing/typevariable/generic_func_parsing_error.c
@@ -1,0 +1,37 @@
+// Tests to make sure generic function call errors are parsed correctly.
+//
+// More specifically, we are testing for below cases of error :
+// 1) Make sure that syntatic error in calling generic functions are handled
+// 2) Using generic function call syntax on normal function will not be
+//    allowed. 
+// 3) a special case that I found. Description below.
+//
+// RUN: %clang_cc1 -fcheckedc-extension -verify %s
+
+_For_any(T) T Foo(T a, T b) {
+  T a;
+  return a;
+}
+
+void Bar() {
+  // There will be parsing error here, since you cannot create a variable of
+  // type void. This results in the Expr representation of Foo to be NULL for
+  // some reason. Since I use isa<DeclRefExpr>() in the code, assert kicks in
+  // saying you cannot do isa<> on NULL pointer, crashing compiler. I fixed the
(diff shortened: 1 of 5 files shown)


tests/parsing/typevariable/generic_func_parsing_error.c
```diff
@@ -0,0 +1,39 @@
...
+//
+// RUN: %clang_cc1 -fcheckedc-extension -verify %s
+
+_For_any(T) T Foo(T a, T b) {
+  T a;
+  return a;
+}
+
+void Bar() {
+  // There will be parsing error here, since you cannot create a variable of
```
dtarditi: This seems pretty specific to the clang implementation.  I'd recommend moving this test to the clang\test\CheckedC directory.

tests/parsing/typevariable/generic_func_parsing_error.c
```diff
@@ -0,0 +1,39 @@
...
+  // saying you cannot do isa<> on NULL pointer, crashing compiler. I fixed the
+  // code to check whether the Expr* is NULL or not.
+  void* x, y; //expected-error{{variable has incomplete type 'void'}}
+  Foo<void*>(x, y);
+  return;
+}
+
+void CallGenericFunction() {
+  void *x, *y;
+  // I was wondering if I should create a separate check and error message to
```
dtarditi: Please don't write comments in the first person.  It is confusing as to later readers as to who "I" was.   It is better to write something like "We could create a separate check and error message to say "Bar is not a generic function, so we cannot have this syntax".  However, it seems sufficient enough for the clang to generate error message as if generic functions were never implemented ..."

dtarditi: The comment is pretty specific to clang.  The Checked C repo should contain language conformance tests, not detailed tests or explanations of clang functionality.  These kinds of tests should go in the Checked C clang repo instead.   It is useful to have a test in this repo of applying type arguments to a non-generic function.  To handle the clang-specific nature of the choice of error messages, I I recommend duplicating the Bar method in the Checked C clang test directory and moving the detailed explanation there.

PR #178, opened 2017-06-29
Add tests for lexical equality of expressions.

dtarditi:
This implements tests for https://github.com/Microsoft/checkedc-clang/issues/107

As part of statically checking bounds declarations, we need to check symbolic equality of expressions and to lexicographically order expressions (for sorting commutative operands).  This implements language-level tests of bounds declarations that indirectly cover lexical equality/inequality.


dtarditi:
As part of statically checking bounds declarations, we need to check symbolic equality of expressions and to lexicographically order expressions (for sorting commutative operands).  This implements language-level tests of bounds declarations that indirectly cover lexical equality/inequality.

PR #190, opened 2017-08-29
Correct Interface Type of free()

lenary:
the `free` function is unique as it has to be polymorphic over all three pointer kinds, unchecked pointers, ptrs, and array_ptrs. 

There are several potential interface types for `free`. They are:
1. `void free(void* p);` No interface type. Doesn't work as no implicit casts between checked ptrs and unchecked ptrs)
1. `void free(void* p : itype(_Ptr<void>));` ptr interface type. Doesn't let you pass array_ptrs to `free`.
1. `void free(void* p : itype(_Array_ptr<void>));` array_ptr interface type. Doesn't let you pass ptrs to `free`, for complex (soon-to-be-relaxed) reasons involving banning casts involving checked pointers to void.
1. `void free(void* p : count(1))` relies on knowing the size of void which is potentially undefined (but 1 byte in clang).
1. `void free(void* p : byte_count(1));` Oddly enough, this is the correct type. When you cast a ptr to an array_ptr, there will be a check (usually static) that the pointer can reasonably point to something, and is in range. This says "yeah, we have to point to something, but it can be as little as 1 byte wide". This still requires casting from `Array_ptr<T>` to `array_ptr<void>`.

This changes free to use `byte_count(1)` in its itype.

### DIFF
--- a/include/stdlib_checked.h
+++ b/include/stdlib_checked.h
@@ -53,7 +53,7 @@ void *aligned_alloc(size_t alignment, size_t size) : byte_count(size);
 void *calloc(size_t nmemb, size_t size) : byte_count(nmemb * size);
 void free(void *pointer : byte_count(1));
 void *malloc(size_t size) : byte_count(size);
-void *realloc(void *pointer  : itype(_Ptr<void>), size_t size) : byte_count(size);
+void *realloc(void *pointer : byte_count(1), size_t size) : byte_count(size);
 
 // TODO: strings
 // char *getenv(const char *n);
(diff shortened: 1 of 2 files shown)


dtarditi:
Choice #2 (bounds-safe interface of Ptr<void>, the existing code) will become synonymous with the Choice #5 (the proposed change), with some upcoming compiler and spec changes.   Choice 5 more clearly communicates the intent, so we should make this change, regardless of the upcoming changes.

lenary:
I have:
- Also improved the type of `realloc` to use `byte_count(1)`.
- Added tests. They currently fail, but every single one of the examples there should pass.

lenary:
I won't commit it until you have a fix to Microsoft/checkedc-clang.

PR #192, opened 2017-08-31
Add variant of array_ptr for pointers to null-terminated arrays.

dtarditi:
This change adds a new type `nt_array_ptr` that is a variant of `array_ptr` for pointers to null-terminated arrays.   The change:
- describes the new type in core-extensions.tex.   
- describes a new checked null-terminated array type go with `nt_array_ptr`
- describes the runtime bounds checking rules for nt_array_ptr.

Most of the rules later in the specification for `array_ptr` apply to `nt_array_ptr`.    I initially started down the route of updating all occurrences of `array_ptr` to say "or `nt_array_ptr`".   I decided that was resulting in lots of changes that didn't help understanding much and made the text longer.   There is a paragraph in the description of `nt_array_ptr` that says that the rules for `array_ptr` usually apply to `nt_array_ptr` and that we mention `nt_array_ptr` only when the rules are different.

This change also cleans up some aspects of the spec:
- Remove the text that says bounds-checking is done for &e, if e is a dereference or subscript operation.  This is incorrect and doesn't match what the implementation does.
- Makes the description of bounds checks more precise, by describing the use of temporaries instead of expressions during bounds checking (the expression being bounds checked is not re-evaluated).
- Revise the section describing the deferral of the evaluation of bounds expression.

Still to do (in another pull request):
- Update the rules for conversions between checked pointer types.
- Update the description of inference of bounds and the checking of bounds declarations to cover the widening of bounds for `nt_array_ptr`.   If we determine the element at the upper bound is not a null-terminator, we can widen the bounds of the `nt_array_ptr` by 1.

### DIFF
--- a/spec/bounds_safety/core-extensions.tex
+++ b/spec/bounds_safety/core-extensions.tex
@@ -52,7 +52,7 @@ lowercase versions of the keywords.  It is assumed
 that \keyword{checkedc.h} is included before examples.
 
 \section{New kinds of pointer types}
-New pointer types are added to C. Each pointer type can be
+Three new checked pointer types are added to C. Each pointer type can be
 used in place of `\texttt{*}':
 \begin{enumerate}
 \item
@@ -75,6 +75,9 @@ used in place of `\texttt{*}':
   bounds are checked at runtime if necessary. Pointer arithmetic can be
   done on these pointer types. The resulting pointers do not need to be
   in bounds.
+\item
+ \ntarrayptrT: this is a variant of \arrayptrT\ for null-terminated
+ arrays.  It is described in Section~\ref{section:nullterm-types}
 \end{enumerate}
 
 Unchecked C pointer types that use `\texttt{*}' and are unchecked in their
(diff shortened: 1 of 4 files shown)


mwhicks1:
I presume no implementation has been done yet, and so you will not be entirely sure of the design until you've done some implementation and experimentation?

spec/bounds_safety/core-extensions.tex
```diff
@@ -259,6 +264,99 @@ \subsection{Propagation and type definitions}
...
+\end{verbatim}
+
+A \keyword{nt\_checked} array with a dimension \var{d} converts
+to an \ntarrayptr\ with \var{d} elements.   This means that programs
+can still read the element containing the null terminator.  However,
+attempting to overwrite it is a runtime error.  The \verb+sizeof+
+operator applied to a null-terminated array includes the null
+terminator in the size computation.
+
+Because \ntarrayptr\ extends \arrayptr, the discussion and rules for
```
lenary: We need a way to make this obvious. I also think it opens up the spec to lots of "except"s, which we have said we want to avoid.

dtarditi:
To answer @mwhicks1 question, there has not been any implementation.   Yes, implementation and experimentation will be required before we are sure of the design.

dtarditi:
@gdr-at-ms, @lenary, @bharadwajy, and I discussed the design yesterday.  Pointers to null-terminated arrays are different than regular array_ptrs because their bounds can be widened based on program actions.    We discussed how to integrate this into the language design.

One possibility is to use a program analysis to infer widened bounds that are subsequently used during runtime bounds checking.  Nobody liked this idea.  It could make the behavior of programs unpredictable.  If the bounds aren't actually widened, a program could fail unexpected at runtime.  The bounds that are used for the check would not be written down in the program, which would make understanding what happened hard.  Note that in the current Checked C design, bounds checking always uses declared bounds.   Any bounds checks are based on expressions written in the program text.

We agreed that a better design choice for bounds inferred via a program analysis is to require that memory accesses using the inferred bounds be provably in bounds at compile time.   Otherwise, the program will be rejected at compile time.   That way there are no unexpected at runtime errors.  

Here is my current design proposal is:
- Programmers can declare bounds for nt_array_ptrs.   In that case, nt_array_ptrs follow the same dynamic bounds checking rules as array_ptrs, except that the memory exactly at the upper bounds can be read.  This is so that we can check for the null-terminator.  The lower bound must also be less than or equal to the address being read.
- When checking bound declarations for nt_array_ptr variables involving bounds, the checking algorithm may usedwider bounds based on program analysis.   However, the wider bounds are not used in dynamic bounds checking.
- If programmers don't declare bounds for an nt_array_ptr variable, the bounds at the declaration will be those for 0-element array  However, wider bounds may be inferred at uses based on program analysis.   Any memory accesses using inferred bounds must be provably in bounds at compile time.

@bharadwajy and I discussed the fact that writes to memory could cause a memory read.   This would cause problems with null-terminated pointers to volatile data.  Even reads can have side effects on hardware with memory-mapped I/O.   The proposed design eliminates this extra memory read.

dtarditi:
I've updated the text to match the proposed design.   @lenary, I also called out clearly that the rules for array_ptr generally hold for nt_array_ptr by elevating the paragraph to a new subsection.  We'll see if this holds up after working through all the details of the changes.

Describing the program analysis to widen bounds is left for future work.  I opened issue #195 to track that.

spec/bounds_safety/core-extensions.tex
```diff
@@ -259,6 +264,108 @@ \subsection{Propagation and type definitions}
...
+% or another \arrayptrT\ value.
+
+Modifiers can be applied to the reference type of
+an \ntarrayptr, in the same way that they can be applied to the
+referent type of an \arrayptr.
+The function \verb+compute_length+
+should not modify the contents of a string. Here
+is a version that uses \verb+const+ to declare that:
+\begin{verbatim}
+int compute_length(nt_array_ptrconst char> s);
```
lenary: missing `<` after `nt_array_ptr`

spec/bounds_safety/variable-bounds.tex
```diff
@@ -602,43 +625,91 @@ \section{Insertion of bounds checks at pointer dereferences}
...
+The check ensures that the memory access is within the bounds of the entire multi-dimensional
+array.  Given \texttt{\var{e1}[\var{e2}][e3]}, the expression \texttt{\var{e1}[\var{e2}]} is
+evaluated.  However, \texttt{\var{e1}[\var{e2}]} has array type, so no lvalue
+conversion is inserted. That means that no bounds check is inserted:
+\texttt{\var{e1}[\var{e2}]} becomes pointer arithmetic.  The bounds check
+is inserted as part of the subscript operation involving \var{e3}.
+
+This means that bounds are not checked for each individual dimension access.
+Accessing outside of the bounds for an
+individual dimension is a violation of the C Standard and logically incorrect,
```
lenary: perhaps clearer as:
> Accessing outside of the bounds for an individual inner dimension is a voilation...

spec/bounds_safety/core-extensions.tex
```diff
@@ -259,6 +264,108 @@ \subsection{Propagation and type definitions}
...
+A problem with using an analysis to determine bounds
+is that the bounds are not written down in the program. 
+This may lead to unexpected failures if the bounds are narrower
+than expected and it may make it difficult to understand
+failures.
+To avoid these problems, we require that memory accesses
+using inferred bounds be provably in-range at compile time.
+To make it provable, pointer arithmetic is limited to
+pointer variables plus constant offsets.  This is enough
+to handle simple cases such as:
```
lenary: I think our specification should not be this limiting, we should require a certain level of static analysis, but also allow for improved static analysis which would presumably allow more programs.

dtarditi:
I've addressed your feedback - I agree the comment about pointer arithmetic involving variables and constants was too limiting.  That's a minimum bar that we should support.

PR #194, opened 2017-09-06
Test returning a checked pointer from a function with an interop type.

dtarditi:
Add tests for returning checked pointers from functions with bounds-safe interfaces for their return values.  Modify an existing test to allow this behavior.   

These tests are for the compiler change in https://github.com/Microsoft/checkedc-clang/pull/368.


tests/typechecking/checked_scope.c
```diff
@@ -1914,3 +1914,73 @@ typedef ptr<int *(void) : itype(ptr<int>)> callback_fn2;
...
+int *f101(int len) : count(len) {
+  _Checked{
+    _Array_ptr<int> p = 0;
+    return p;
+  }
+  return 0;
+}
+
+
+// Return an _Array_ptr from a checked scope for a function with a
```
lenary: This comment is incorrect.

Issue #207, opened 2017-10-13
Describe typing rules for string literals.


dtarditi:
In the C specification, string literals have the type "array of character type", where there are several possible character types.  In Checked C, that would correspond to the type "unchecked array of character type". 

We want to allow string literals to be used in checked scopes and where checked pointers are expected.  We extend the typing rules for string literals:
- In a checked scope,  a string literal will have the type "checked array of character".
- When a string literal occurs where a checked pointer type is expected, the type will be altered to be a checked pointer type.   We modify the rule for conversion for checked array types to pointer types.  If a string literal with an unchecked array type occurs in a context where a checked pointer type is expected, the type of the expression is altered to be "checked pointer type of character."  The kind of the checked pointer type shall match the kind of the expected type.  The contexts where this happens are the right-hand side of an assignment, an argument to a  function call, or in an initializer.  The string literal may be possible parenthesized.

dtarditi:
This can be generalized to array literals:

When a string literal or compound literal with array type occurs where a checked pointer type is expected, the type will be altered to be a checked pointer type.   We modify the rule for conversion for checked array types to pointer types.  If a string literal or compound literal expression with type "array of T" occurs in a context where a checked pointer type is expected, the type of the expression is altered to be "checked pointer type to T."  The kind of the checked pointer type shall match the kind of the expected type.   The contexts where this happens are the right-hand side of an assignment, an argument to a  function call, or in an initializer.  
    - The string or array literal may be possible parenthesized.
    - The type "checked pointer to T must be a valid type - null-terminated array pointers to non-scalar and non-integer types are not allowed.

dtarditi:
This work is complete.

PR #208, opened 2017-10-17
Test strings with checked types and initialized checked data.

dtarditi:
This change corresponds to a matching compiler [change](https://github.com/Microsoft/checkedc-clang/pull/396) that adds support for string literals with checked types and completes support for initialized checked data.   The language rules for string and array literals are discussed in issue #207 (they haven't been added to the Checked C specification yet).

The compiler change also improves error messages for types that occur syntactically in expressions and are used in checked scopes.  Those types must be checked and everything they use must be checked too.   This change updates checked_scope_basic.c to use new error messages.

Testing:
- Passes automated testing with the updated compiler.

### DIFF
--- a/tests/static_checking/initializers.c
+++ b/tests/static_checking/initializers.c
@@ -88,7 +88,7 @@ void f4(void) checked {
   nt_array_ptr<char> t37 = g37[1];
 
 
-  f3("abc");   // expected-error {{passing 'char checked[4]' to parameter of incompatible type 'char *'}}
+  f3("abc");   // expected-error {{passing 'char nt_checked[4]' to parameter of incompatible type 'char *'}}
   unchecked{
     f3("abc");
   }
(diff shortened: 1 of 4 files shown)


tests/static_checking/initializers.c
```diff
@@ -24,3 +23,122 @@ extern void f() {
...
+int g21 nt_checked[] = { 0, 1, 2, 0 };
+char g22 checked[3]nt_checked[3] = { { 0, 1, 0}, { 1, 1, 0}, { 3, 1, 0 } };
+char g23 checked[3]nt_checked[4] = { "abc", "def", "fgh" };
+
+//
+// Checked pointers with initialized array literals
+//
+
+nt_array_ptr<int> g30 : count(4) = (int[]) { 0, [2] = 2, 3, 5, 0 };
+nt_array_ptr<int> g31 : count(0)  = (int _Checked[]) { 0, [2] = 2, 3, 5, 0};
```
lenary: Is there a good reason to be mixing the `nt_array_ptr` style identifiers and the `_Checked` style keywords? Surely if we have the former we have the `#defines` from `stdchecked.h`

awruef: I share this nit.

dtarditi: I agree that we shouldn't mix styles for keywords.   The mixed style was the result of copying small snippets from another file.  I'll fix this.

PR #211, opened 2017-10-23
Test bounds checking for null-terminated pointers.

dtarditi:
Checked C allows reads at exactly the upper bound for null-terminated pointers
so that the potential terminating value can be examined.

This change adds new dynamic tests that check  that the the memory location at
the upper bound is accessible for null-terminated array pointers, and not
accessible for non-null-terminated array pointers.

### DIFF
--- a/tests/dynamic_checking/bounds/nullterm_pointers.c
+++ b/tests/dynamic_checking/bounds/nullterm_pointers.c
@@ -1,0 +1,199 @@
+// Test bounds checking in checked scopes of uses of pointers
+// and arrays with bounds-safe interfaces.
+//
+// RUN: %clang -fcheckedc-extension %s -o %t1 -Werror -Wno-unused-value
+// RUN:  %t1 1 | FileCheck %s --check-prefixes=CHECK,NO-BOUNDS-FAILURES-1
+// RUN:  %t1 2 | FileCheck %s --check-prefixes=CHECK
+// RUN:  %t1 3 | FileCheck %s --check-prefixes=CHECK
+// RUN:  %t1 4 | FileCheck %s --check-prefixes=CHECK
+//
+// RUN:  %t1 21 | FileCheck %s --check-prefixes=CHECK,NO-BOUNDS-FAILURES-2
+// RUN:  %t1 22 | FileCheck %s --check-prefixes=CHECK
+// RUN:  %t1 23 | FileCheck %s --check-prefixes=CHECK
+// RUN:  %t1 24 | FileCheck %s --check-prefixes=CHECK
+
+#include <assert.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <stdchecked.h>
+
+struct CountedNullTermString {
+  nt_array_ptr<char> s : count(len);
+  int len;
+};
+
+struct CountedString {
+  array_ptr<char> s : count(len);
+  int len;
+};
+
+int test1(void);
+int test2(void);
+void test3(void);
+void test4(void);
+
+int test21(struct CountedNullTermString *p);
+int test22(struct CountedString *p);
+int test23(struct CountedNullTermString *p);
+int test24(struct CountedString *p);
+
+
+// Handle an out-of-bounds reference by immediately exiting. This causes
+// some output to be missing.
+void handle_error(int err) {
+  _Exit(0);
+}
+
+
+
+
+// This signature for main is exactly what we want here,
+// it also means any uses of argv[i] are checked too!
+int main(int argc, array_ptr<char*> argv : count(argc)) {
+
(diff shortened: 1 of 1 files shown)


tests/dynamic_checking/bounds/nullterm_pointers.c
```diff
@@ -0,0 +1,199 @@
...
+void handle_error(int err) {
+  _Exit(0);
+}
+
+
+
+
+// This signature for main is exactly what we want here,
+// it also means any uses of argv[i] are checked too!
+int main(int argc, array_ptr<char*> argv : count(argc)) {
```
lenary: Low Priority comment, no need to change anything yet, just thought I could remark.

These could now probably be `int main(int argc, array_ptr<nt_array_ptr<char>> argv : count(argc)) {...` This might need more work on our checked headers though.

tests/dynamic_checking/bounds/nullterm_pointers.c
```diff
@@ -0,0 +1,199 @@
...
+    return EXIT_FAILURE;
+  }
+
+  if (argc < 2) {
+    // CHECK-NOT: Requires Argument
+    puts("Requires Argument");
+    return EXIT_FAILURE;
+  }
+
+  struct CountedNullTermString nullterm = { "abcde", 5 };
```
lenary: My first thought was "Shouldn't `len` include the null terminator in this case, so have a value of 6?"

But then I got confused, as `nullterm.s[5]` is `'\0'` and this is what we're looking for. I think I have confusion around what the `count(n)` refers to, and how this interfaces with the size of the string literal array, which in this case I believe is `char nt_checked[6]` or am I wrong? Evidently I don't have my mental model entirely correct yet and need to go back to your recent revisions of the spec.

tests/dynamic_checking/bounds/nullterm_pointers.c
```diff
@@ -0,0 +1,199 @@
...
+  return i;
+}
+
+void test3(void) {
+  nt_array_ptr<char> s : count(0) = "hello";
+  while (*s) {
+    *s = 'd';
+  }
+  // CHECK-NOT: expected bounds failure on write
+  puts("Expected bounds failure on write");
```
lenary: The case of the printed string and the expected string need to match, I believe.

tests/dynamic_checking/bounds/nullterm_pointers.c
```diff
@@ -0,0 +1,199 @@
...
+      puts("Unexpected test case");
+      return EXIT_FAILURE;
+  }
+  return EXIT_SUCCESS;
+}
+
+// Read exactly at the upper bound of a null-terminated pointer.
+// Should not cause a runtime fault.
+int test1(void) {
+  nt_array_ptr<char> s : count(0) = "hello";
```
lenary: This is the case we want our static analysis to definitely catch, right?

tests/dynamic_checking/bounds/nullterm_pointers.c
dtarditi: We're not quite there yet: we need to treat nt_array_ptr<char> as having count(0), which isn't what will happen right now.

tests/dynamic_checking/bounds/nullterm_pointers.c
dtarditi: A null-terminated array with n elements has bounds of count(n-1).   We exclude the null terminator from the count.   The count includes only array elements that we can read and write.   

We can read the nth element (which is the null terminator).  However, given that we count from zero, that means that we can read a[n], when a is null-terminated pointer.

tests/dynamic_checking/bounds/nullterm_pointers.c
dtarditi: I'm not sure what you mean by "catch".  Do you mean "determine it will always succeed and eliminate the dynamic check"?   Or do you mean (for the other cases), "determine it will always fail at run time and warn about that?

tests/dynamic_checking/bounds/nullterm_pointers.c
dtarditi: Good catch!  I originally had longer strings for the NOT cases and decided that was a bad idea, because it would be easy to miss typos.

dtarditi:
I've fixed the casing typo.

tests/dynamic_checking/bounds/nullterm_pointers.c
lenary: I meant “is this the case where we always want our static analysis to increase the inferred count, so that we can access all the way to the end of the null-terminated array?” which seems to be correct.

dtarditi: Yes, this is the case where we want to increase the inferred count.

PR #213, opened 2017-10-24
Rename bounds(none) to bounds(unknown).

dtarditi:
This change goes with a matching compiler change.   Update the Checked C tests to match the specification.

### DIFF
--- a/tests/parsing/declaration_bounds.c
+++ b/tests/parsing/declaration_bounds.c
@@ -15,7 +15,7 @@ extern array_ptr<int> c : count(cLen);
 extern array_ptr<int> d : byte_count(20);
 extern array_ptr<int> e : byte_count(5 * sizeof(int));
 extern array_ptr<int> f : byte_count(cLen * sizeof(int));
-extern array_ptr<int> g : bounds(none);
+extern array_ptr<int> g : bounds(unknown);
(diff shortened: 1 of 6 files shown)


tests/parsing/declaration_bounds.c
```diff
@@ -80,11 +80,11 @@ extern void f3(array_ptr<int> arr : count(5)) {
...
   array_ptr<int> p : count(count) = arr;
-  int none = 0;
-  // none is a contextual a keyword if it follows 'bounds' '('
-  array_ptr<int> q : bounds(none) = arr;
-  array_ptr<int> r : bounds(none + arr, arr + len) = arr; // expected-error {{expected ')'}} expected-note {{to match this '('}}
-  array_ptr<int> s : bounds(arr + none, arr + len) = arr;
+  int unknown = 0;
+  // unknown is a contextual a keyword if it follows 'bounds' '('
+  array_ptr<int> q : bounds(unknown) = arr;
+  array_ptr<int> r : bounds(unknown + arr, arr + len) = arr; // expected-error {{expected ')'}} expected-note {{to match this '('}}
```
lenary: We could (later) improve the parser here not to error (and instead to backtrack), but this doesn't matter that much.

PR #219, opened 2017-11-01
Add examples of using null-terminated strings.

dtarditi:
Here are some examples of using null-terminated strings adapted from K&R.   The code illustrates some basic aspects of using strings, as well as the sometimes trickier aspects of writing to a string.

I plan to clean up the code once the compiler change for making count(0) be the default bounds for nt_array_ptr is complete.

### DIFF
--- a/samples/string-helpers.c
+++ b/samples/string-helpers.c
@@ -1,0 +1,215 @@
+// This file contains examples that illustrate the use of strings in
+// Checked C.  Strings are null-terminated arrays of characters in C.
+// The examples are adapted from "The C Programming Language", Second Edition,
+// by Brian Kernighan and Dennis Ritchie.
+//
+// To compile the file using clang, on Unix/Mac use
+//  clang -o string-helpers -fcheckedc-extension string-helper.c
+// On Windows use:
+//  clang -o string-helpers.exe -fcheckedc-extension string-helper.c
+// 
+// Then run the program with 6 string arguments, the 3rd of which should
+// be an integer  For example:
+//   string-helpers hello meet -125 fed brown red
+//  
+// Overview of using null-terminated arrays in Checked C:
+//
+//   In Checked C, pointers to null-terminated data are represented using the
+//   nt_array_ptr type.  Arrays of null-terminated data are represented
+//   using the nt_checked type.
+//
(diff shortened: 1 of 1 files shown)


mwhicks1:
These examples compile for me and are helpful. Thanks!

One question: I see that you included the following headers for this example:
```
#include <ctype.h>
#include <stdchecked.h>
#include <stdio_checked.h>
```
Does this mean that the checked versions of stdio.h, stdlib.h etc. are not automatically chosen by the compiler? Should I be changing references to, say, `stdio.h` in code that I'm porting to be `stdio_checked.h` ?

mwhicks1:
One thing I am confused about, of the examples:
```
checked int my_strlen(nt_array_ptr<char> p : count(0)) {
  int i = 0;
  nt_array_ptr<char> s : count(i) = p; /* 1 */
  while (s[i] != '\0')
    ++i; /* 2 */
  return i;
}
```
I don't understand why this works. Statement `/* 1 */` implies that `s` has bounds `s` to `s+i`. Does statement `/* 2 */` effectively expand the bounds on `s`, which is justified by the NUL check that precedes it?

mwhicks1:
One more question, this time about the last example. The type of `main` is as follows:
```
int main(int argc, nt_array_ptr<char> argv[] : count(argc)) {
```
Here, the `count` refers to `argv`, not the strings it contains. But you write `argv` with type `[]` and not as an `array_ptr` --- why is this OK? I checked that the following works, so I am wondering about the difference.
```
int main(int argc, array_ptr<nt_array_ptr<char>> argv : count(argc)) {
```

dtarditi:
To answer Mike's questions:
1. Yes, you should change references to headers to use the checked versions of them when available.  The compiler is not automatically choosing them.
2. Yes, statement 2 widens the bounds at the ++i.   This is justified by the preceding null check.
3. Good catch - the [] should be checked[].   Otherwise we end up with a bounds-safe interface.   Note that if the body of main is in  a checked scope, this doesn't matter.  Anything with a bounds-safe interface is treated as thought it were checked.

In some cases, I modified the original code K&R code to change `char []` to `nt_array_ptr<char>`.   I could have changed it to use zt_checked[] instead.

PR #220, opened 2017-11-02
Test default bounds for null-terminated array pointers

dtarditi:
Null-terminated array pointers have a default bounds of count(0).   This adds tests of bounds checking and type checking for null-terminated array pointers with programmer-specified bounds and default bounds.   This change matches a corresponding clang compiler change (https://github.com/Microsoft/checkedc-clang/pull/406).

There need to be additional tests added of struct members with default bounds of count(0).   I'd like to commit the compiler change soon to unblock people, so I'm opening this PR now, even though further work is needed.


mwhicks1:
Fine with me. I'm not attempting to run anything right now, but will be soon. So feel free to get this in there.

dtarditi:
Thanks - I've opened #224 to track adding more tests here.

PR #222, opened 2017-11-02
Add another example of using strings and add string support to stdio_checked.h

dtarditi:
The example is adapted from a simple "pattern finding" program from "The C Programming Language, Second Edition", by Kernighan and Ritchie.  The example uses only checked code.    The example does output using fputs; I updated fputs and the rest of stdio_checked.h with annotations for strings.

This example requires the compiler changes in https://github.com/Microsoft/checkedc-clang/pull/406 to compile.  I had to fix two bugs, which I'll add to the compiler pull request soon.

### DIFF
--- a/samples/find-pattern.c
+++ b/samples/find-pattern.c
@@ -1,0 +1,112 @@
+// This file contains an example that illustrates the use of strings in
+// Checked C.  The example is adapted from "The C Programming Language,
+// Second Edition", by Brian Kernighan and Dennis Ritchie, p. 69.
+// It reads a series of lines and check whether a string occurs in a line.
+// If it does, it prints the line.
+//
+// To compile the file using clang, on Unix/Mac use
+//  clang -o find-pattern -fcheckedc-extension find-pattern.c
+// On Windows use:
+//  clang -o find-pattern.exe -fcheckedc-extension find-pattern.c
+//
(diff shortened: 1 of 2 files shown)


samples/find-pattern.c
```diff
@@ -0,0 +1,112 @@
...
+// exceeding the maximum number of characters allowed in a line.
+#define MAXLINE 50
+
+#pragma BOUNDS_CHECKED ON
+
+int getline(char line checked[] : count(max), int max);
+int strindex(char source nt_checked[], char searchfor nt_checked[]);
+void print_string(char str nt_checked[]);
+
+int main(int argc, nt_array_ptr<char> argv checked[] : count(argc)) {
```
lenary: I must have missed where we can now support bounds on VLAs, which is excellent. I should go back and review some code that uses them.

dtarditi: We only added support for specifying bounds on incomplete array types.   There's some overlap with C's VLA feature because incomplete types with bounds can be used in some of the places where VLAs can be used.  We haven't done the broader work to integrate with C's VLA feature, though.

lenary: So I can't employ the same trick with a VLA at the end of a struct yet?

PR #238, opened 2017-11-13
Compiler now warns about some out-of-bounds memory accesses.

dtarditi:
Update dynamic tests of runtime bounds checking to suppress these warnings.  We don't expect to see warnings for these tests and turn warnings into errors.

For bounds declaration checking, turn TODO's for expected warnings into pattern matches for the expected warnings.


dtarditi:
Update dynamic tests of runtime bounds checking to suppress these warnings.  We don't expect to see warnings for these tests and turn warnings into errors.

For bounds declaration checking, turn TODO's for expected warnings into pattern matches for the expected warnings.

PR #239, opened 2017-11-15
Make Fortify-source definitions more fine-grained

lenary:
This is an annoying change, but sometimes, despite having `_FORTIFY_SOURCE` set to non-zero, the stdlib won't always give a definition for that a symbol that should be fortified (looking at you, Mac OS X 10.13 and unprotected `fprintf`)

This check introduces per-function guards. They define the normal kind of the function if:
- fortify_source == 0; or
- there is no macro for the function.
If fortify_source is zero (ie, we're not using the fortified versions), and the macro is defined, then we need to undefine it, as it's constant folding definition or similar, and we can fall back on the original definition.

Damnit I wish that system headers were more cooperative.

This will need checking on linux please @awruef 

Testing:
- Passes on Mac OS X 10.13.

### DIFF
--- a/include/stdio_checked.h
+++ b/include/stdio_checked.h
@@ -50,7 +50,10 @@ int setvbuf(FILE * restrict stream : itype(restrict _Ptr<FILE>),
 // * Any pointer arguments may not meet the requirements of the
 //  format string.
 //
-#if _FORTIFY_SOURCE == 0
+
+
+#if _FORTIFY_SOURCE == 0 || !defined(fprintf)
+#undef fprintf
 _Unchecked
 int fprintf(FILE * restrict stream : itype(restrict _Ptr<FILE>),
             const char * restrict format : itype(restrict _Nt_array_ptr<const char>), ...);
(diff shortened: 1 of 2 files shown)


include/stdio_checked.h
```diff
@@ -60,15 +63,16 @@ _Unchecked
 int fscanf(FILE * restrict stream : itype(restrict _Ptr<FILE>),
            const char * restrict format : itype(restrict _Nt_array_ptr<const char>), ...);
 
-#if _FORTIFY_SOURCE == 0
+#if _FORTIFY_SOURCE == 0 || !defined(printf)
```
dtarditi: Do you need to undef printf here?

include/stdio_checked.h
```diff
@@ -50,7 +50,10 @@ int setvbuf(FILE * restrict stream : itype(restrict _Ptr<FILE>),
 // * Any pointer arguments may not meet the requirements of the
 //  format string.
 //
-#if _FORTIFY_SOURCE == 0
+
+
+#if _FORTIFY_SOURCE == 0 || !defined(fprintf)
```
dtarditi: Could you add comments to the source code explaining the pattern that you are using here, so that others can understand it later?

include/stdio_checked.h
lenary: Yes I do.

include/stdio_checked.h
lenary: Yes, Sure. It's convoluted because the conditions are complicated.

Issue #235, opened 2017-11-07
nul-terminated array bounds checks


mwhicks1:
I think I'm triggering some bounds checks. Two questions:

1) The bounds check failure is manifesting as SIGILL, "illegal instruction". How should this be manifesting, instead?
```
% ./a.out 1 2 2 33 4 345 34 3 453 43545
Illegal instruction: 4
```
2) It seems that the rules could be relaxed to allow not just reading the last position but also writing a `'\0'` there. Any reason not to allow that? That's what's causing my problem right now.

dtarditi:
I'll address question 2 first.  I don't want my response to get buried  I have more say about 1 than you might expect.

The main cost will be that bounds checks for writes to null-terminated arrays will get a little more expensive in terms of code size.   Given bounds of `bounds(lb, ub)` and a pointer `p`, the bounds checks for writing a value 'v' to a null-terminated array is `lb <= p < ub`   It would be `lb <= p < ub || (v = 0 && p == ub).`     It won't affect code speed much because the common case is still covered by the first clause.  The second clause covers an exceptional case.

I think this will be a pretty common error to make.  I made it in the sample code that I wrote and you made it, and I'm sure lots of other programmers will make it.  We could try to statically detect when a value is being written at the upper bound and warn the programmer that they need to insert a check.  Of course, a static check can still miss cases. 

So the arguments for and against relaxing the rules:
- For: this will be a common error and there won't be a speed effect in the common case.
- Against: there's a potential code size increase.  Programmers can always add explicit checks to avoid writing null at the last location, and we can warn when they seem to have missed a check in common cases.

For 1, this is what's expected right now.   We're using the LLVM trap instruction to signal a runtime fault and that's the behavior.  Eventually I think we should produce a SIGSEGV.    Is the use of SIGILL causing problems with debugging or not getting core dumps?

This issue of how to trap efficiently and give programmers information for debugging is a little more complicated than it seems.   We do introduce lots of trap instructions.   In optimized code they get combined right now, but it isn't clear that you want that behavior in production systems.  When they are combined, you don't know where in a function the buffer overrun happened, which makes debugging challenging.

[...]

PR #242, opened 2017-11-28
Allow writing a 0 at the upper bound of a null-terminated array.

dtarditi:
This change addresses the suggestion from @mwhicks1 in issue #235 that we relax the bounds checking for null-terminated arrays to allow a write of 0 at the upper bound of a null-terminated array.  There will be a matching pull request in the checked-clang repo for compiler change.  

- Update the description of bounds checking in the Checked C specification.
- Add tests to nullterm_pointers.c that check this behavior.
-  Update the sample code in string-helpers.c to delete some special-case code in the "squeeze" function.  This function removes all occurrences of a character from a string.  If no characters are removed, it redundantly overwrites the existing null-terminator.  Before, this had to be special-cased to avoid an out-of-bounds write.   I also updated the sample code to avoid declaring default bounds of count(0) and clarified the comments.

I ran into a subtlety when revising the specification.  When we allow a 0 to be written at the upper bound, we have to expand the allowed range of memory to include the memory at the upper bound.  It is possible for even this expanded range to be empty.   The specification is careful to not allow a write of 0 in this situation., by including a comparison of the pointer address against the lower bound.

### DIFF
--- a/samples/string-helpers.c
+++ b/samples/string-helpers.c
@@ -18,21 +18,23 @@
 //   nt_array_ptr type.  Arrays of null-terminated data are represented
 //   using the nt_checked type.
 //
-//   An nt_array_ptr is like an arraY_ptr, except that it is followed by a
-//   a sequence of elements that is terminated by a null element.
+//   An nt_array_ptr is like an array_ptr, except that the array that it
+//   points to is followed by a sequence of elements that is null-terminated.
+//   The bounds on the nt_array_ptr delimit the array.  The sequence begins
+//   at the upper bound.
 //
-//   These types follow slightly different rules than array_ptr and checked
-//   arrays do:
+//   Because of this,  nt_array_ptr and nt_checked arrays follow slightly
+//   different rules than array_ptr and checked arrays do:
 //   - Given an nt_array_ptr, the element exactly at the upper bound of the
 //     nt_array_ptr can be read (it is the first element of the sequence).
 //     A 0 can be written there using an assignment (there must always be
 //     enough space in the sequence for at least the null terminator).  No
 //     other writes to that element are allowed.
-//   - Given an nt_checked array, it convert to an nt_array_ptr whose count
+//   - Given an nt_checked array, it converts to an nt_array_ptr whose count
 //     excludes the  null terminator (whose count is one less than the declared
 //     size).
 //   - nt_array_ptrs have a default bounds of count(0). This means that you
-//     can can dereference an nt_array_ptr that does not have a bounds
+//     can dereference an nt_array_ptr that does not have a bounds
 //     declaration.
 //
 //   Some simple examples:
--- a/samples/string-helpers.c
+++ b/samples/string-helpers.c
@@ -6,7 +6,7 @@
 // To compile the file using clang, on Unix/Mac use
 //  clang -o string-helpers -fcheckedc-extension string-helper.c
 // On Windows use:
-//  clang -o string-helpers.exe -fcheckedc-extension string-helper.c
+//  clang -o string-helpers.exe -fcheckedc-extension string-helpers.c
 // 
 // Then run the program with 6 string arguments, the 3rd of which should
 // be an integer  For example:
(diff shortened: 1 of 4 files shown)


mwhicks1:
Updated spec seems fine, as you describe it.

samples/string-helpers.c
```diff
@@ -18,41 +18,53 @@
...
-//     upper bound of the nt_array_ptr can be read but not
-//     written.
-//   - Given an nt_checked array, it has count of
-//      of elements that excludes the null terminator (is one
-//      less than the declared size).
-//   - nt_array_ptr have a default bounds of count(0). This
-//     means that you can can dereference an nt_array_ptr without
-//     bounds (NOTE: the compiler is currently not yet inserting
-//     the default bounds.  You'll need to add them explicitly.
+//   An nt_array_ptr is like an arraY_ptr, except that it is followed by a
```
lenary: Typo `array_ptr`

spec/bounds_safety/variable-bounds.tex
```diff
@@ -669,13 +669,33 @@ \section{Insertion of bounds checks at pointer dereferences}
 given \texttt{*\var{e1}} where {\var{e1}} has {\bounds{\var{e2}}{\var{e3}}},
 the compiler computes \var{e1} to some temporary \var{t}.
 The compiler inserts a runtime check that  \texttt{\var{e2} <= \var{t} \&\&
-\var{t} <= \var{e3}}.   For memory writes, the compiler inserts the
-same check as for \arrayptr:
-\texttt{\var{e2} <= \var{t} \&\& \var{t} < \var{e3}}.
+\var{t} <= \var{e3}}.   For memory writes, assignment of 0 (the null value)
+using the assignment operator is allowed at exactly the upper bound.
+The upper bound is the beginning of the sequence and there must be enough
```
lenary: > the upper bound is the beginning of the sequence

This doesn't quite make sense, this is only the case with `count(0)`. Say I have `count(5)` because I know there are 5 non-zero elements, the upper bound is not the beginning of the sequence of elements. Please may you clarify this sentence?

samples/string-helpers.c
```diff
@@ -18,41 +18,53 @@
...
+//   a sequence of elements that is terminated by a null element.
+//
+//   These types follow slightly different rules than array_ptr and checked
+//   arrays do:
+//   - Given an nt_array_ptr, the element exactly at the upper bound of the
+//     nt_array_ptr can be read (it is the first element of the sequence).
```
lenary: > first element of the sequence

Also doesn't quite make sense, see note below. Clarification would help.

samples/string-helpers.c
dtarditi: Fixed.

samples/string-helpers.c
dtarditi: I've clarified the wording of this comment and the specification.    An nt_array_ptr points to an array that is followed by a sequence of elements that is null-terminated.   The bounds on the nt_array_ptr delimit the array.  The sequence begins at the upper bound.

dtarditi:
Thanks for the feedback!.  I've updated the pull request to address it

nmeum:
Hey, I just wanted to inform you about some difficulties we're having in our Checked C project due to this changes. We do understand why this changes was made, we just wanted to give you some feedback on it.

We have a setup were want to make it possible to compile all the code we converted to Checked C with a legacy C compiler (due to various reason I won't explain in detail here). We achieved this through heavy use of bounds-safe interfaces and custom CPP macros which disable Checked C keywords when a legacy C compiler is being used.

The following property causes some problems with code that should compile with both the Checked C and the Legacy C compiler:

> Given an nt_checked array, it converts to an nt_array_ptr whose count excludes the null terminator (whose count is one less than the declared size).

We have a few cases in our legacy C source code where arrays are initialized using `memset(3)` as follows:

```C
char a[5];

/* … */

char *p = a;
memset(p, '\0', 5);

/* … */

/* write some stuff to the array */

/* … */
```

the pointer `p` is later passed to function from `string.h`, e.g. `strcmp(3)` and thus need to be null terminated. In Checked C we would therefore use an `_Nt_checked` type for `a` and an `_Nt_array_ptr` type for `p`. However, the `memset` invocation causes a compiler error then because the bounds for `p` are `count(4)` instead of `count(5)`.

Changing the `memset`invocation to `memset(p, '\0', 4);` would of cause work but causes issues when the code is compiled with a legacy C compiler since the char in the array isn't initialized then.

So we are forced to change all `memset` invocations (like the one above) in our code base to the following:

```C
memset(p, '\0', 4);
p[4] = 0; // Should work in Checked C?
```

Other workaround are possible as well. The point is that any kind of workaround makes it more difficult to retain our Legacy C and Checked C interoperability. We do understand though, that we have a very special Checked C use case and just wanted to give you some feedback on the changes introduced here. It would be easier for us, if `_Nt_checked` arrays would convert to `_Nt_array_ptr` with bounds including the null terminator and runtime checks for ensuring that the null terminator at the upper bound is never overwritten.

**Note:** `p[4] = 0` should work if I understood the spec changes commited here and [the test5() function](https://github.com/Microsoft/checkedc/pull/242/files#diff-6d35565b178c9d20e508f72a1a8ad4b4R221) correctly but somehow causes a compiler error with `clang version 6.0.0 (https://github.com/Microsoft/checkedc-clang c4c385d41c53d68ddcefaba90f82f636fe80e948) (https://github.com/Microsoft/checkedc-llvm d8d78c7a13472b344e429d1ae320bc4f01ecb6db)`. Is this intended?

dtarditi:
@nmeum, thank you for the feedback!    We understand the general feedback, which is that narrowing the bounds by one for an `nt_array_ptr` causes problems using general-purpose C standard library functions such as `memset` and `memcpy` that would otherwise be correct.   We put the null terminator "out of-reach" to avoid problems where someone does something like `memset(p, 'a', 5)`. 

I'll open an issue tracking this.   It can be difficult to add runtime checking. We'd need to differentiate between pointers that started as array_ptr vs. those that started as nt_array_ptr.  We do not want not change data layout, so we have no way of knowing that solely at runtime.  There are a few other approaches we could take:
- Special case uses of well-known C library functions for zeroing.  This is not general, but would be pretty straightforward to do.
- A slightly more general approach would be to allow a programmer to declare that a function only writes zeros, if a specific argument is 0.  We could then special case the handling of `nt_array_ptr`.
- Enhance the static reasoning system so that we can convert an `nt_array_ptr` to an `array_ptr` with restrictions on writing to the final character. This would be challenging to do because we'd need a a vocabulary for talking about the restrictions, as well as a way of reasoning that the restrictions are satisfied at uses of the pointer.

A workaround that might require fewer source changes to maintain is to overallocate the` _Nt_checked` array by 1 character.  Of course, this can be problematic if this is for an existing data structure that must fit in a certain amount of memory or where widening could have memory usage effects (for example, widening from 128 characters to 129 characters might cause some memory allocators that use powers of 2 blocks to allocate 256 bytes, which would be significant increase in memory usage).

PR #247, opened 2018-01-08
Error messages for problems involving return statements. 

dtarditi:
C programs can have problems with return statements that would result in undefined or incorrect program behavior. C compilers typically warn about these problems (in other languages, these would typically be compile-time errors because of the severity of the problem).  For Checked C, these problems need to result in errors when they occur in a checked scope or involve bounds or check pointer types.  This
change adds tests for these problems.


dtarditi:
@awruef, I committed this change.  Please feel free to provide feedback later.

PR #249, opened 2018-01-22
Checked C typing rules for literals.  Also describe handling of initializers.

dtarditi:
Describe extensions to C typing rules that allow literals to be used in checked scopes or where expressions with checked pointers are expected.   This completes the work for issue #207.   For checked scopes, string literals are given the type "null-terminated checked array of character", where "character" is the character type for the string literal.   For contexts where an expression with a checked pointer type is expected but not found, convert string literals and compound array literals to checked pointer types.

Also describe the rules for initializers.   The types of initializers are determined by the  type of the variable being declared, so there isn't any change need for checked C.  

Add examples.

Move the section describing checked scopes earlier in Chapter 2, so that it precedes the discussion of literals (this is most of the change diff shown on GitHub).

### DIFF
--- a/spec/bounds_safety/core-extensions.tex
+++ b/spec/bounds_safety/core-extensions.tex
@@ -730,19 +730,42 @@ ptr<int> p = 0;
 int arr1 checked[5] = { 0, 1, 2, 3, 4 };
 int arr2 checked[3] = { 0, 1, 2 };
 
-struct Pair {
+// Initialize pointers to checked arrays
+array_ptr<int> ap1 = arr1;
+array_ptr<int> ap2 = (int checked[4]) { 0, 0, 0, 0 };
+
+// Initialize a multi-dimensional checked array
+float unit_transform checked[2][2] = { { 1.0f, 0.0f}, { 0.0f, 1.0f }};
+
+// A ragged 2-dimensional array, where the inner dimension size varies.
+struct inner_array {
+  int len;
+  array_ptr<double> elems; // bounds declaration missing because we
+                           // haven't described them yet.
+};
+
+typedef struct inner_array ragged_2d_arr[];
+
+ragged_2d_arr data = 
+  { { 3, (double checked[]) { 1.0, 0.0, 0.0 } },
+    { 1, (double checked[]) { 5.0 } },
+    { 0, 0 },
+    { 2, (double checked[]) { 6.0, 7.0 } } 
+  };
+
+struct BufferPair {
   int len1;
   array_ptr<int> buf1;
   int len2;
   array_ptr<int> buf2;
 };
 
-// Initialize all members of an instance of Pair.
-struct Pair p1 = { 5, arr1, 3, arr2 };
+// Initialize all members.
+struct BufferPair p1 = { 5, arr1, 3, arr2 };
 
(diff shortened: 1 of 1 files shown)


spec/bounds_safety/core-extensions.tex
```diff
@@ -470,6 +470,359 @@ \subsection{Variable and member initialization}
...
+
+On the other hand,  declarations in checked scopes can use unchecked pointer
+types and unchecked array types, provided that the declarations provide a
+bounds-safe interface.   These are described in
+Section~\ref{section:function-bounds-safe-interfaces}.
+Variables and functions used in checked scopes are 
+allowed to have or use checked pointer types, checked array types, or
+unchecked pointer types and unchecked array types with bounds declarations.
+
+A new pragma directive \texttt{BOUNDS\_CHECKED} is introduced to control whether
```
lenary: I don't think this is designed well enough to put into the specification yet. I have a proposal to email around, using `push`/`pop` so you can nest these pragma scopes (particularly useful for headers etc)

spec/bounds_safety/core-extensions.tex
```diff
@@ -470,6 +470,359 @@ \subsection{Variable and member initialization}
...
+    ...
+}
+\end{verbatim}
+
+When a function call occurs in a checked block, the function being
+called does not have to be declared as checked. The notion of whether a
+scope is checked or not checked is lexical and the function definition
+is a separate lexical scope.
+
+C allows declarations of functions without prototypes, where the types
```
lenary: Might be worth calling out explicitly (or in a footnote) that these are sometimes called "K&R Declarations".

spec/bounds_safety/core-extensions.tex
```diff
@@ -470,6 +470,359 @@ \subsection{Variable and member initialization}
...
+The brace-enclosed list may be a partial list and specify values only for
+part of an object or array.  In that case, the remaining part of the
+array or object is initialized with default values (0 for integers or
+pointers).
+
+For Checked C, no modifications are needed to initializers.
+the value of the initializer is interpreted based on the type of the
+declared variable and its members.  A checked array or pointer is initialized
+the same way that an unchecked array or pointer is initialized.  For
+null-terminated checked arrays, we require that
```
lenary: what do we require?

spec/bounds_safety/core-extensions.tex
dtarditi: This is just existing text that I moved around.  I agree that it still needs further work and will happy to see your proposal to use `push / pop`.

spec/bounds_safety/core-extensions.tex
dtarditi: Good suggestion - I'll add a footnote for this.

spec/bounds_safety/core-extensions.tex
dtarditi: I'll fill in the missing text.

spec/bounds_safety/core-extensions.tex
lenary: Ah, ok, I just read the added column in split view. This comment is not a blocker on the PR.

dtarditi:
It is ready for further review.  Thanks!

Issue #193, opened 2017-09-06
Describe type checking of return statements for bounds-safe interfaces


dtarditi:
Section 5.1.4 covers typechecking in the presence of bounds-safe interfaces.  It is silent on the typechecking of return statements for functions with return bounds-safe interfaces.   These should be treated the same away that assignments to parameters/global variables with bounds-safe interfaces are treated.   The spec should mention this.

dtarditi:
PR #250 completed this work.

PR #250, opened 2018-01-22
Return statements with bounds-safe interfaces.

dtarditi:
Update the specification to cover type checking and bounds checking of return statements with bounds-safe interfaces.  This addresses issue #193.


dtarditi:
@lenary, more small changes to look at, if you have time.

Issue #206, opened 2017-10-11
Describe type checking rules for conditional expressions.


dtarditi:
We have implemented type checking rules in the compiler for conditional expressions where one or both branches produce values with checked pointer types.  The specification needs to be updated to describe them.  The implemented rules try to choose the most specific type for the conditional expression, which is the type that provides the most guarantees.

Given `e1 ? e2 : e3` and the types for each of the arms, we use the following rules in the compiler.  The rules require that both branches produce pointers to T.
1. If one branch has unchecked pointer type and the other branch is an array_ptr or ptr, use the checked type as the type of the entire expression.   An implicit cast is inserted for the unchecked pointer-typed branch.   For array_ptrs, the bounds will be limited to what is true for both branches.  For ptrs, the unchecked pointer-typed branch must have known bounds.
2. If one branch has unchecked pointer type or array_ptr type and the other branch is an nt_array_ptr, use array_ptr as the type of the entire expression.   The bounds will be limited to the intersection of the bounds for both branches, which will prevent the null terminator from being overwritten.

We probably want to implement a rule related to ptrs and array_ptrs:
3. If one branch has ptr type and other branch has array_ptr or nt_array_ptr type, use the ptr type as the type of the entire expression.  This will lead to an implicit conversion being inserted for the non-ptr typed branch.  This will mean that we have to be able to prove statically that the resulting value for that branch meets the bounds requirements for a ptr.

dtarditi:
The specification has been updated.

Issue #150, opened 2017-04-20
Allow implicit conversions between checked pointers to T


dtarditi:
We should allow ptr<T> to be implicitly converted to array_ptr<T> and vice versa.   Checking of bounds declarations will enforce that bounds constraints are met.    This implicit conversion rule was partially implemented in checked-clang change list 2fb1a6. I think the case for conditional expressions was missed, and there may be other cases that were missed.

This work item is to update the spec and tests, add tests if necessary, and make sure that the clang implementation is complete.

dtarditi:
The specification is updated.  I've opened a separate work item for the clang implementation for conditional expressions (https://github.com/Microsoft/checkedc-clang/issues/444)

PR #251, opened 2018-01-25
Update rules for implicit conversions.

dtarditi:
This updates the type checking rules for implicit casts. This addresses issues 
#150 and #206 n the specification.  It removes the subsections for each kind
of cast between pointer kinds and presents a unified framework instead.
- Use a diagram to describe what kinds of pointers can be converted to other
  kinds of pointers.  It is easier to visualize this than to explain in text.
  The conversions form a lattice. We can use this fact to more simply
  describe conversion rules.
- Allow conversions from checked void pointers.  We allow this by necessity
  already in the implementation.  It is needed so that malloc
  can be used in checked code.
- Mention that we plan to replace casts from checked void pointers.
- Move the description of assignment compatibility earlier.

In the description of C-style casts, remove the special case for ptr<void>.
It would be confusing to expand it now that we allow conversions from
array_ptr<void too. State in text instead that the conversions between
checked void pointers are allowed.

Add a description of the type checking rules for conditional expressions.
This was only partly covered, and that description was incomplete.
It is easier to explain this more generally.  Note that the compiler doesn't
quite match the spec.  It doen't allow casts between ptr<T> and array_ptr<T>,
which are now allowed.  I opened Checked C clang issue 444 to cover this.

Update some tests for type checking of conditional expressions to more
precisely check for arm types of the form 'T *' and "checked pointer to void"
that we're not merging them to T *.  This could then be converted to checked
pointer type.  The existing tests would pass and things would look OK, even
though from a typing perspective, it wouldn't be.


dtarditi:
@lenary, if you have a chance, could you review the specification changes?

spec/bounds_safety/interoperation.tex
```diff
@@ -299,42 +303,48 @@ \subsection{Non-examples}
...
+\end{center}
+\caption{Implicit conversions between different kinds of pointers}
+\label{fig:implicit-type-conversion}
+\end{figure}
+
+Each pointer type has a referent type (the type to which the pointer refers).
+For example, \arrayptrT\ points to \var{T}.   For an implicit conversion, the
+source referent type and the destination referent type must be the same type or
+be {\em assignment compatible}.  Informally, the destination referent type must
+be at least as checked as the source referent typing.  Losing checking is not allowed.
```
lenary: > Losing Checking is not allowed

You then have a point below, based on the referent type D and the referent type S:
> D and S are both checked arrays, both unchecked arrays, or D is a checked array and S is an unchecked array, and

So in this case you lose "inner" checked-ness of D, but this is fine because you really have `<pointer type>D` to `<pointer type S>` and you don't lose the "outer" checked-ness of these two pointer types, you only lose the "inner" checked-ness of the arrays. The next bullet point says the arrays are constant length anyway, so all code is actually fine, right?

spec/bounds_safety/interoperation.tex
```diff
@@ -344,35 +354,74 @@ \subsection{From unchecked pointers to checked pointers}
...
+There are several important points about these conversions:
+\begin{itemize}
+\item  While an implicit conversion may allow typechecking to succeed,
+       checking of bounds declarations may still fail.  Implicit
+       conversions are treated as though they are explicit C cast
+       operations during the checking of bounds declarations.
+\item  Implicit conversions from checked pointer types to
+      unchecked pointer types are not allowed because that would cause a loss of bounds safety.
+      There is an important exception: they are allowed at bounds-safe interfaces
+      bounds-safe interfaces, where an lvalue is treated as having both an unchecked
```
lenary: duplication of "bounds-safe interfaces"

spec/bounds_safety/interoperation.tex
```diff
@@ -299,42 +303,48 @@ \subsection{Non-examples}
...
+kind conversions that are allowed.  Any kind can be converted to another
+kind that is reachable following arrows. For example, it makes sense to allow an unchecked
+pointer type to be converted to an \arrayptr\ or \ptr.
+
+\begin{figure}
+\begin{center}
+\begin{tikzpicture}[sibling distance=0.5in]
+\node (ptr) {\ptr};
+\node (arrayptr)   [above=of ptr]      {\arrayptr}    edge[<->, thick] (ptr);
+\node (ntarrayptr) [right=of arrayptr] {\ntarrayptr}  edge[->, thick] (arrayptr);
```
lenary: might be more obvious if you do "[right=of unchecked]" and reorder so that works. This would get you something where the LUB/GLB of the lattice was easier to "see"

dtarditi: Thanks for pointing out that it is non-obvious that this is a lattice.  In fact, it isn't a lattice!   I've reworked the section on conditional expressions to use a slightly different diagram that is a semi-lattice (defined GLB, but not a LUB).

spec/bounds_safety/interoperation.tex
dtarditi: Yes, that is correct.

spec/bounds_safety/interoperation.tex
dtarditi: Fixed.

Issue #98, opened 2017-01-24
Update initialization requirements for automatic variables


dtarditi:
For now, automatic variables with `_Array_ptr` or `_Ptr` type must have initializers.  The spec needs to be updated to describe this clearly.

dtarditi:
This work is complete.

PR #252, opened 2018-01-31
Update description of initialization requirements

dtarditi:
Variables with checked pointer type that may be used to access memory must be initialized when they are declared.  This also applies to variables with members or elements with checked pointer type that may be used to access memory.  Update the specification to make this clear.   Include some examples.  This addresses issue #98.

The compiler already checks these requirements for scalar variables.  I opened https://github.com/Microsoft/checkedc-clang/issues/445 to track checking these requirements for struct and array variables.

Add tests for the struct/array cases to the Checked C tests.


dtarditi:
@awruef or @lenary, could you take a quick look at this update to the specification, if you have time?

PR #263, opened 2018-03-16
Stack Object Protection with Low Fat Pointers, NDSS17

lcavallaro:
Added reference to "Stack Object Protection with Low Fat Pointers", NDSS 2017


lcavallaro:
Added reference to "Stack Object Protection with Low Fat Pointers", NDSS 2017

PR #267, opened 2018-03-21
Update to new syntax for bounds_cast expressions.

dtarditi:
Update tests and examples to the new syntax for bounds cast expressions.

We changed bounds cast expressions to take a bounds expression as the
second argument, instead of implicitly inferring the type of bounds
expression from the number of arguments.  This is more verbose but
clearer.  Bounds cast expressions now have two forms:

Op<T>(e1)
Op<T>(e1, bounds-expression)

where Op is one of _Assume_bounds_cast or _Dynamic_bounds_cast
and T must be a pointer type.


dtarditi:
This is a pretty straightforward update, so I'm going to commit now.  @awruef, let me know if you see any issues to address after the commit.

PR #268, opened 2018-03-22
Test bounds checking of interop types with bounds expressions.

dtarditi:
We recently added compiler support for declaring bounds-safe interfaces that have interop types and bounds expressions.  Add tests of runtime bounds checking involving parameters declared this way.   The variables are declared in unchecked scopes and used in checked scopes.   Add tests for these combinations:

- checked array interop types.  This implies bounds that are the count of the first dimension of the array.
- array_ptr interop types with bounds expressions.
- nt_array_ptr interop types with bounds expressions.

We already had a test of nt_array_ptr with no bounds expression, which implies bounds of count(0).


dtarditi:
I'm going to commit this now.  @awruef, let me know if you have suggestions or comments and I'll address them.

Issue #272, opened 2018-04-06
Conversion of checked array types to pointer types


nmeum:
Since you seem to appreciate feedback about the Checked C extension itself we wanted to inform you about an issue we were having with the conversation rules for checked arrays which made it more difficult for us to convert our code base to Checked C.

The conversation rules for checked Arrays are defined in Section 2.5.2 of version `v0.7` of the Checked C specification as follows:

> If the type of an expression is “checked array of T”, the type of the expression is altered to be array_ptr<T>. If the type of an expression is “checked null-terminated array of T ”, the type of the expression is altered to be nt_array_ptr<T>.
    
We didn't want to convert everything at once and thus used a one-module-at-a-time approach which causes some issues with the conversion rule cited above.

For instance, assume you have the following struct definition in a header file:

```C
struct Foo {
  char a[23];
}
```

When converting the header file defining this struct you would convert it as follows:

```C
struct Foo {
  char a _Checked[23];
}
```

However, this may cause other (not yet converted) parts of your software to no longer compile. One example for Legacy C code that would no longer compile with the new struct definition is:

```C
void myfunc(struct Foo *foo) {
  char *p = foo->a; // needs to be changed to (char*)foo->a
}
```

This behaviour is explained in Section 5.2 of the Checked C specification:

> Implicit conversions from checked pointer types to unchecked pointer types are not allowed because that would cause a loss of bounds safety.

The code above can, however, be fixed by using an explicit cast to an unsafe pointer type. It would be neccessary though to find and changes all Legacy C parts of the codebase where the `a` member of `struct Foo` is used as a pointer, which makes conversation more difficult in our experience.

You don't have this problem with `struct members` of a pointer type because you can use bounds-safe interfaces for those. We are therefore wondering if it would be possible to introduce a 'bounds-safe checked array type' or to make the conversation rules for checked Array types more relaxed to allow **implicit** conversations of Checked Arrays to unsafe pointer types in unchecked program scope? This would definitly increase interoperability with Legacy C code parts.

dtarditi:
Thank you for your thoughtful feedback.  I think we already have a solution for this issue.  You can use "bounds-safe interface types" to provide an alternate checked type for arrays:
```
struct Foo {
   char a[23] : itype(char _Checked[23]);
};
```
Interface types allow you to declare a "more checked" version of a type to be used in a checked scopes.  They were originally added to handle things like:
```
struct S {
   int **p;
}
```
[...]

Issue #269, opened 2018-03-22
Update spec to allow an interop type and a bounds expression.


dtarditi:
We now allow bounds-safe interfaces to have both interop types and bounds expression, to handle cases like `int **p`.    This is implemented in the compiler.  The spec needs to be updated to describe this.

dtarditi:
This is complete.

Issue #100, opened 2017-01-26
Update motivation and definitions of Explicit Dynamic Checks


lenary:
After a discussion today, we realise we need to update section 2.8 on dynamic checks to reflect the motivation that these should be able to be checked statically or dynamically, as the compiler wishes, and therefore should contain only non-modifying expressions.

dtarditi:
This section has been updated.

PR #283, opened 2018-05-02
Update specification to match implementation.

dtarditi:
I plan to release an updated version of the spec because the BOUNDS_CHECKED pragma has been renamed to CHECKED_SCOPE.  Address some other issues where the specification doesn't match the implementation.
- Address issue #100 (update motivation and definitions of explicit dynamic check).  Add a statement that for dynamic_check(e), e must be nonmodifying.
- Address issue #269.  Update specification to allow an interop type and a bounds expression as part of a bounds-safe interface.
- Change the wording to be clear that bounds-safe interface can be applied to unchecked arrays too.  This addresses some feedback in issue #272.
- Remove the requirement that every parameter for a function or every member of a struct/member must be annotated with a bounds-safe interface, if one of them is annotated.  The checking as implemented actually works on a per parameter/per member basis.  Requiring that everything be converted could get in the way of incremental conversion.

### DIFF
--- a/spec/bounds_safety/core-extensions.tex
+++ b/spec/bounds_safety/core-extensions.tex
@@ -1152,8 +1152,7 @@ for these reasons:
 \end{itemize}
 
 The expression being checked must be non-modifying because
-\code{dynamic_check} exists to allow programmers to express and check
-program invariants.  An invariant is a statement about program state and
+\code{dynamic_check} checks if an invariant holds.  An invariant
 should not modify program state.
 
 The following example illustrates why having an escape hatch from static
--- a/spec/bounds_safety/core-extensions.tex
+++ b/spec/bounds_safety/core-extensions.tex
@@ -445,6 +445,7 @@ of type \var{T}, the following rules apply:
 \end{itemize}
 
 \subsection{Rules for conversion of checked array types to pointer types}
+\label{section:array-to-pointer-conversion}
 
 If the type of an expression is ``checked array of
 \var{T}'', the type of the expression is altered to be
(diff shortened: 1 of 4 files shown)


spec/bounds_safety/core-extensions.tex
```diff
@@ -1149,6 +1151,11 @@ \section{Programmer-inserted dynamic checks}
   control-flow of the program.
 \end{itemize}
 
+The expression being checked must be non-modifying because
```
awruef: Maybe re-phrase: "dynamic_check checks if an invariant holds, these invariants should not modify program state." and maybe put it in the previous paragraph? The two sentences just seem a little awkward.

spec/bounds_safety/interoperation.tex
```diff
@@ -810,6 +835,56 @@ \subsection{Examples}
...
+\end{lstlisting}
+
+\subsection{Bounds expression and type annotations}
+A bounds-safe interface declaration can have both a bounds expression and a type
+annotation.  This allows programmers to declare bounds-safe interfaces for declarations involving
+nested pointer types.
+
+Consider a bounds-safe interface for \lstinline{char **p}.
+The checked type for the variable can be declared using a type annotation and
+the bounds for the first level of pointer can be declared using a bounds expression:
```
awruef: Is this really the bounds for the first level of pointer? ptr<ptr<char>> is a mult-level pointer value. I think what you're saying is that the checked type is specified starting at the outer-most pointer type declaration and working inward, right?

dtarditi:
Thanks for the feedback - I think I've addressed it.

PR #287, opened 2018-06-12
add initializers for initialization required variables, and some test cases for array/struct/union unit testing.

sxl463:
Once we finished the initialization checking for array/struct/union variables with checked pointer elements/members, we should update the test cases also. Here is what we did:
 
(1) add initialization directly 
e.g. ptr<int(int, int)> arr7 checked[5] : count(5) = {0};

(2) for some complex struct declared with pragma CHECKED_SCOPE, we add an expected-error comment 

We also add some test cases for array/struct/union unit testing in  /static_checking/initializers.c

### DIFF
--- a/tests/static_checking/initializers.c
+++ b/tests/static_checking/initializers.c
@@ -198,7 +198,7 @@ void f6(void) {
     float y;
   };
 
-  struct checked_value_no_bounds init_S2 = { 0 }; // no bounds; initializer not required, should pass
+  struct checked_value_no_bounds init_S2; // no bounds; initializer not required, should pass
   struct checked_value_no_bounds uninit_S2; // no bounds; initializer not required, should pass
 
   ptr<struct checked_value_no_bounds> arr_init   checked[20] = { 0 }; // initialized arry, should pass
@@ -206,8 +206,8 @@ void f6(void) {
 
   struct checked_value_has_bounds {
     int x;
-    array_ptr<char> lower :count(5);
-    array_ptr<char> upper :count(5);
+    array_ptr<char> lower : count(5);
+    array_ptr<char> upper : count(5);
     float y;
   };
 
@@ -227,13 +227,13 @@ void f6(void) {
     array_ptr<int> upper;
     float y;
   };
-  union u_checked_value_no_bounds init_U = { 0 }; // no bounds; initializer not required, should pass
+  union u_checked_value_no_bounds init_U; // no bounds; initializer not required, should pass
   union u_checked_value_no_bounds uninit_U; // no bounds; initializer not required, should pass
 
   union u_checked_value_has_bounds {
     int x;
-    array_ptr<char> lower :count(5);
-    array_ptr<char> upper :count(5);
+    array_ptr<char> lower : count(5);
+    array_ptr<char> upper : count(5);
     float y;
   };
   union u_checked_value_has_bounds init_U2 = { 0 }; // has bounds; initializer required and we did, should pass
@@ -250,9 +250,9 @@ void f6(void) {
 
   typedef struct {
     int data;
-    array_ptr<char> name :count(20);
+    array_ptr<char> name : count(20);
     struct Node* next;
-  }Node;
+  } Node;
 
   Node n_err; // expected-error {{containing a checked pointer must have an initializer}}
   Node n = { 0 };
@@ -261,7 +261,7 @@ void f6(void) {
   typedef struct {
     struct Range r;
     Node center;
-  }Circle;
+  } Circle;
 
   Circle C_err; // expected-error {{containing a checked pointer must have an initializer}}
   Circle C = { 0 };
@@ -269,7 +269,7 @@ void f6(void) {
   typedef struct {
     Circle Outer;
     Circle Inner;
-  }Annulus;
+  } Annulus;
 
   Annulus a_err; // expected-error {{containing a checked pointer must have an initializer}}
   Annulus a = { 0 };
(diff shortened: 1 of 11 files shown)


tests/static_checking/initializers.c
```diff
@@ -179,4 +177,185 @@ void f6(void) {
...
+  struct has_uninitialized_ptr_member uninit_S;   // expected-error {{containing a checked pointer must have an initializer}}
+
+  struct checked_value_no_bounds {
+    int x;
+    array_ptr<int> lower;
+    array_ptr<int> upper;
+    float y;
+  };
+
+  struct checked_value_no_bounds init_S2 = { 0 }; // no bounds; initializer not required, should pass
```
dtarditi: I think you need to remove the initializer here.

tests/static_checking/initializers.c
```diff
@@ -179,4 +177,185 @@ void f6(void) {
...
+  struct struct_with_checked_field_has_bounds init_nested_S = { 0 }; // has bounds, initializer required and we did, should pass
+  struct struct_with_checked_field_has_bounds uninit_nested_S; // expected-error {{containing a checked pointer must have an initializer}}
+
+  union u_checked_value_no_bounds {
+    int x;
+    array_ptr<int> lower;
+    array_ptr<int> upper;
+    float y;
+  };
+  union u_checked_value_no_bounds init_U = { 0 }; // no bounds; initializer not required, should pass
```
dtarditi: I think you need to remove the initializer.

tests/static_checking/initializers.c
```diff
@@ -179,4 +177,185 @@ void f6(void) {
...
+  };
+  struct struct_with_checked_union_field_has_bounds init_SU = { 0 }; // has bounds, initializer required and we did, should pass
+  struct struct_with_checked_union_field_has_bounds uninit_SU; // expected-error {{containing a checked pointer must have an initializer}}
+
+
+  typedef struct {
+    int data;
+    array_ptr<char> name :count(20);
+    struct Node* next;
+  }Node;
```
dtarditi: Could you add a space after the `}`?  There are a few other places where you need to do this.

tests/static_checking/initializers.c
```diff
@@ -179,4 +177,185 @@ void f6(void) {
...
+    array_ptr<int> lower;
+    array_ptr<int> upper;
+    float y;
+  };
+  union u_checked_value_no_bounds init_U = { 0 }; // no bounds; initializer not required, should pass
+  union u_checked_value_no_bounds uninit_U; // no bounds; initializer not required, should pass
+
+  union u_checked_value_has_bounds {
+    int x;
+    array_ptr<char> lower :count(5);
```
dtarditi: Could you add a spaceafter the ':'?   There are a few other places where you need to do this.

tests/static_checking/initializers.c
```diff
@@ -179,4 +177,185 @@ void f6(void) {
...
+  }Node;
+
+  Node n_err; // expected-error {{containing a checked pointer must have an initializer}}
+  Node n = { 0 };
+
+
+  typedef struct {
+    struct Range r;
+    Node center;
+  }Circle;
```
dtarditi: Could you add a space after the `}`?

tests/static_checking/initializers.c
```diff
@@ -179,4 +177,185 @@ void f6(void) {
...
+  int x : count(5); // expected-error {{with a bounds expression must have an initializer}} expected-error {{have a pointer or array type}}
+  int y : count(6) = 0; // expected-error {{have a pointer or array type}}
+}
+
+// integer with bounds expression needs to be checked
+void f8 (void) {
+
+  typedef struct {
+    int a;
+    int b : count(10);  // expected-error {{have a pointer or array type}}
```
dtarditi: I think you should a different kind of bounds expression, as count bounds expressions trigger an error.  How about bounds((array_ptr<int>) i, (array_ptr<int>) i + 10)?

tests/static_checking/initializers.c
```diff
@@ -179,4 +177,185 @@ void f6(void) {
...
+
+// integer with bounds expression needs to be checked
+void f8 (void) {
+
+  typedef struct {
+    int a;
+    int b : count(10);  // expected-error {{have a pointer or array type}} 
+  }S;
+  
+  S s1; // expected-error {{containing a checked pointer must have an initializer}}
```
dtarditi: This error message seems confusing, as S doesn't have a checked pointer.  It has an integer.

tests/typechecking/checked_arrays.c
```diff
@@ -241,11 +241,11 @@ extern void check_dimensions1(void) {
 ('dim_checked' is a checked array)}}
 
   // checked parenthesized declarators
-  int (t6 checked[10])[10];            // checked propagates to enclosing array declarators
+  int(t6 checked[10])[10];            // checked propagates to enclosing array declarators
```
dtarditi: Could you undo these formatting changes?

tests/static_checking/initializers.c
sxl463: bounds case added now

sxl463: The new kind test case added now

tests/static_checking/initializers.c
```diff
@@ -179,4 +177,197 @@ void f6(void) {
...
+  int x : count(5); // expected-error {{with a bounds expression must have an initializer}} expected-error {{have a pointer or array type}}
+  int y : count(6) = 0; // expected-error {{have a pointer or array type}}
+}
+
+// integer with bounds expression needs to be checked
+void f8 (void) {
+  // for bounds expr kind like "bounds((array_ptr) i, (array_ptr) i + 10)"
+  typedef struct {
+    int a;
+    int b : bounds( (array_ptr<int>) b, (array_ptr<int>) b + 10); // expected-warning 2 {{cast to '_Array_ptr<int>' from smaller integer type 'int'}}
```
dtarditi: I believe this will break on a 32-bit targets.   Specifically, the compiler won't issue a warning there.  I suggest you change the type of `b` to `unsigned long long`

sxl463: Yes I failed on DevTest X86 windows testing. Can you explain it further?

dtarditi: The size in bits of integers and pointers can vary depending on the target platform.  On a 64-bit system, `int` is often a 32 bits.  Pointers will typically be 64-bits. On a 32-bit system, they will both be 32-bits.  So you won't get the warning on a 32-bit system.   See https://docs.microsoft.com/en-us/windows/desktop/WinProg64/migration-tips for some discussion of 32 bit vs. 64-bit platform issues.

Issue #291, opened 2018-07-21
Bounds-safe interface for strncmp needs to be revised


dtarditi:
@AnnaKornfeldSimpson was converting code for JSON parsing to use Checked C.  She found that the current bounds-safe interface for `strncmp` is overly strict for null-terminated pointers:
```
int strncmp(const char *src : count(n), const char *s2 : count(n), size_t n);
```
The definition for `strncmp` is that it compares two possibly null-terminated strings up to n characters.  This bounds-safe interface says that the strings must have at least n characters, which is too strict for null-terminated strings.

A nice thing about bounds-safe interfaces and checked types is that they are opt-in.  My suggestion is that we give `strncmp` a bounds-safe interface where it takes `nt_array_ptr` values as arguments.  A I'm guessing that's the most common usage, but it may not be.  We should then define extended" inline versions of `strncmp` that take `array_ptr`s or a combination of `nt_array_ptr`' and `array_ptr`    For example:
```

int strncmp(const char *src : itype(nt_array_ptr), const char *s2 : itype(nt_array_ptr), size_t n);
extern inline int strncmp_array_ptr(const char *src : count(n), const char *s2 : count(n), size_t n) {
  _Unchecked { return strncmp(src, s2, n); 
}
```
 This disambiguates the use cases.  As part of opting-in, the programmer or a tool can choose which one to use.  We could put these extended versions in a separate opt-in header file or we could add them to the existing header file for now.

AnnaKornfeldSimpson:
It might also be helpful to have a **note:** produced by the compiler when it has errors/warnings proving `strncmp` (and those of other functions that have extended versions) that mentions these alternate versions and how to enable them. Is there an easy way to do that?

Issue #290, opened 2018-07-20
Bounds-safe interface for free incompatible with nt_array_ptr


AnnaKornfeldSimpson:
The bounds-safe interface for the stdlib function `free` in `stdlib_checked.h` assumes that the argument has bounds that include at least one byte of memory:
`void free(void *pointer : byte_count(1));`

However, we define `nt_array_ptr` to have bounds of `count(0)` by default (and then do bounds widening), which means that `nt_array_ptr` type memory cannot be sent to `free`.

Since developers free strings very regularly, forcing them to add something extra to that call in the case of `nt_array_ptr` seems both confusing and bothersome. Can we do something under the hood to support calls to `free` with `nt_array_ptr` args, without losing the bounds checking we have for `array_ptr`?

awruef:
It seems like `free` could be `byte_count(0)`? Any reason why not? Sounds to me like "`free` takes a pointer to some object where we don't really know anything about the size (and if we did it wouldn't matter because we don't communicate the size to `free` anyway)"?

dtarditi:
We can either change the bounds-safe interface to count(0) or add an alternate inline function definition of free for _nt_array_ptr.  I'm leaning toward adding an alternate definition for nt_array_ptr.     A nice thing about bounds-safe interfaces is that they are opt-in, so it is OK to suggest that calls be changed.  Note that we have a problem with needing an alternate bound-safe interface for strncmp too (issue #291). 

We could place the alternate definition in an opt-in header file or include it in the existing header file.

AnnaKornfeldSimpson:
In yesterday's meeting, we discussed changing the bounds-safe interface for `free`. I think @Prabhuk was checking to make sure it was ok to free 0 bytes?

dtarditi:
Currently we have a bounds-safe interface for `free` of byte_count(1).   @Prabhuk checked the C11 Specification.  It is possible to allocate zero-sized objects, which means you it is possible to free them.  From section 7.22.3 of the C11 Specification about `malloc` and other allocation functions:

> If the size of the space requested is zero, the behavior is implementation-deﬁned: either a null pointer is returned, or the behavior is as if the size were some nonzero value, except that the returned pointer shall not be used to access an object.

I think we should change the bounds-safe interface for `free` to be `byte_count(0)`.

We did consider having a bounds-safe interface for `free` that didn't have have any bounds:
```
void free(void * : itype(_Array_ptr<void>));
```

[...]

PR #292, opened 2018-07-24
New header with additional bounds-safe interfaces

AnnaKornfeldSimpson:
Added a new header `checkedc_extensions.h` which specifies extra bounds-safe interfaces that can be chosen by the programmer when functions have different interfaces for `array_ptr` and `nt_array_ptr`. Added it to the `CMakeLists.txt` in the include folder and added it to the header inclusion test `tests/typechecking/redeclare_libraries.c`. Passes this test and functions properly when compiling example third-party code.

Addresses #291 and #290

### DIFF
--- a/include/checkedc_extensions.h
+++ b/include/checkedc_extensions.h
@@ -10,7 +10,8 @@
 #include <stdlib_checked.h>
 #include <string_checked.h>
 
-// default strncmp assumes nt_array_ptr; this option is for array_ptr
+// default strncmp has a bounds-safe interface nt_array_ptr;
+// this option is for array_ptr
 extern inline int strncmp_array_ptr(const char *src : count(n), const char *s2 : count(n), size_t n) {
   _Unchecked { return strncmp(src, s2, n); }
 }
--- a/include/checkedc_extensions.h
+++ b/include/checkedc_extensions.h
@@ -1,0 +1,24 @@
+//--------------------------------------------------------------------------//
+// Additional bounds-safe interface options for functions that have         //
+// different interfaces for array_ptr and nt_array_ptr args. Programmer can //
+// choose to use these over the default bounds-safe interfaces.             //
+//////////////////////////////////////////////////////////////////////////////
+
+#ifndef __CHECKED_C_EXTENSIONS_H
+#define __CHECKED_C_EXTENSIONS_H
+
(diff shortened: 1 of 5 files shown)


include/string_checked.h
```diff
@@ -109,7 +109,9 @@ int strcoll(const char *src1 : itype(_Nt_array_ptr<const char>),
             const char *src2 : itype(_Nt_array_ptr<const  char>));
 
 
-int strncmp(const char *src : count(n), const char *s2 : count(n), size_t n);
+int strncmp(const char *src : itype(_Nt_array_ptr<const char>),
```
dtarditi: It would be good to point to the extended version here and to discuss strncmp's behavior: that it takes a possibly null-terminated string as arguments and checks up to n characters.   We can't overload functions based bounds-safe interfaces.  A simple way to encode this in Checked C is to have two separate functions with different bounds-safe interfaces, one for _Nt_array_ptr and another for _Array_ptr.

include/stdlib_checked.h
```diff
@@ -56,7 +56,7 @@ unsigned long long int strtoull(const char * restrict nptr :
 // TODO: express alignment constraints once where clauses have been added.
 void *aligned_alloc(size_t alignment, size_t size) : byte_count(size);
 void *calloc(size_t nmemb, size_t size) : byte_count(nmemb * size);
-void free(void *pointer : byte_count(1));
+void free(void *pointer : byte_count(1)); // for _Array_ptr
```
dtarditi: I believe this can actually be used for _Array_ptr or _Ptr.  It'd be good point out here that there's a separate version of free for _Nt_array_ptr.

include/checkedc_extensions.h
```diff
@@ -0,0 +1,24 @@
...
+// choose to use these over the default bounds-safe interfaces.             //
+//////////////////////////////////////////////////////////////////////////////
+
+#ifndef __CHECKED_C_EXTENSIONS_H
+#define __CHECKED_C_EXTENSIONS_H
+
+#include <stdlib_checked.h>
+#include <string_checked.h>
+
+// default strncmp assumes nt_array_ptr; this option is for array_ptr
```
dtarditi: Instead of saying "assume", I'd suggest "has a bounds-safe interface for nt_array_ptr."

PR #294, opened 2018-07-25
Bounds-safe interface for function in errno

AnnaKornfeldSimpson:
One of the files I'm converting uses `errno`. This makes CheckedC unhappy because the function that makes errno (inside a layer of macros) returns an `int*`. This PR adds a new `errno_checked.h` header file to provide a bounds-safe interface for that function. It passes the `redeclare_libraries` test and gets rid of the errors in my converted file. `errno.h` is a short header; I believe this is the only function.

### DIFF
--- a/include/CMakeLists.txt
+++ b/include/CMakeLists.txt
@@ -7,7 +7,8 @@
 # library just to get the Checked C include wrpaper files.
 # So we place the header files with the compiler.
 
-set(files 
+set(files
+  errno_checked.h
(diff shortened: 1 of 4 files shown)


AnnaKornfeldSimpson:
Added a very simple test that triggers the bounds-checking.
Also added a bounds-safe interface for Windows.

AnnaKornfeldSimpson:
Fixed the build configuration.

PR #298, opened 2018-08-02
Require byte_count(0) of the arg to free

AnnaKornfeldSimpson:
We decided to change the bounds-safe interface requirement of the argument passed to free to be byte_count(0) rather than byte_count(1). This means we no longer need the extra version of free in checkedc_extensions, or several lines of commentary.

Addresses #290


AnnaKornfeldSimpson:
Passes checkedc tests.

PR #299, opened 2018-08-08
Bounds-safe interface for assert

AnnaKornfeldSimpson:
Both Windows and Linux have some interesting things going on with their use of assert. Add a new header with bounds-safe interface for the function eventually called by assert macros in the case of failure.


include/assert_checked.h
```diff
@@ -0,0 +1,32 @@
+//---------------------------------------------------------------------//
+// Bounds-safe interfaces for functions in stdlib.h that               //
```
dtarditi: This comment needs to be updated.

PR #303, opened 2018-08-11
Adding support for polymorphic bounds safe interface functions.

Prabhuk:
Parsing logic of _For_any and _Itype_for_any seems to fail in linux.
Merging this PR must wait until the root cause is found and fixed.

### DIFF
--- a/tests/typechecking/itype_generic_functions.c
+++ b/tests/typechecking/itype_generic_functions.c
@@ -44,7 +44,7 @@ void CallItypeGenericFunctions(void) {
   checked {
     p = validItypeGenericFunction<float>(5, p2, p2);
     //Checked scope expects type arguments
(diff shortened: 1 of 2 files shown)


dtarditi:
The new tests for https://github.com/Microsoft/checkedc-clang/pull/546 look good.

Could you add tests under tests\runtime_operations that actually define and call generic functions?   The tests should actually run code and cover the various cases, including, definitions of generic functions, definitions of itype generic functions, and redeclarations of existing functions to have a generic bounds-safe interface type.

I think we need to expand the testing of generic functions.  Could you open an issue assigned to you to:
1. Test arguments that are pointers to generic functions.
2. Test struct members that are pointers to generic function types.
3. Test returning a pointer to a  generic function.
4. Test features 1-3 embedded in a generic functions.

Thanks!

PR #306, opened 2018-08-20
Polymorphic bounds safe interface stdlib functions.

Prabhuk:
Changed the signature of malloc, calloc and free to make them polymorphic bounds safe interface functions. 

Testing:
1. Regression tests are modified to have type parameters in calls to malloc, calloc and free
2. LNT tests are updated to have type parameters in calls to malloc, calloc and free


AnnaKornfeldSimpson:
This one looks good to me!

PR #313, opened 2018-09-02
Add POSIX socket header

AnnaKornfeldSimpson:
Adds bounds-safe interfaces for functions in the POSIX sys/socket.h header.
@Prabhuk can you verify that I'm using `Itype_for_any()` correctly in this header?


include/socket_checked.h
```diff
@@ -0,0 +1,124 @@
...
+#ifdef __USE_GNU
+extern int recvmmsg (
+    int __fd, 
+    struct mmsghdr *__vmessages : itype(_Array_ptr<struct mmsghdr>) count(__vlen), 
+    unsigned int __vlen, 
+    int __flags, 
+    struct timespec *__tmo : itype(_Ptr<struct timespec>));
+#endif
+
+_Itype_for_any(T) extern int getsockopt (
```
dtarditi: I think you should have getsockopt and setsockopt be non-generic functions.  Really this is a giant union of values with optname as the union tag.

PR #315, opened 2018-09-05
Making Memcpy, bsearch, realloc  polymorphic generic functions

Prabhuk:
Signatures of memcpy, bsearch and realloc are modified to make them polymorphic bounds safe interface functions

### DIFF
--- a/include/stdlib_checked.h
+++ b/include/stdlib_checked.h
@@ -78,12 +78,12 @@ int system(const char *s : itype(_Nt_array_ptr<const char>));
 // on parameters based on size.  Currently we are requiring that
 // bounds in parameters lists be closed with respect to variables
 // in the parameter list.
-_Itype_for_any(T) void *bsearch(const void *key : itype(_Array_ptr<const T>) byte_count(size),
-              const void *base : itype(_Array_ptr<const T>) byte_count(nmemb * size),
-              size_t nmemb, size_t size,
-              int ((*compar)(const void *, const void *)) :
-                itype(_Ptr<int(_Ptr<const T>, _Ptr<const T>)>)) :
-                byte_count(size);
+_Itype_for_any(T) void *bsearch(const void *key : itype(_Ptr<const T>),
+                                const void *base : itype(_Array_ptr<const T>) byte_count(nmemb * size),
+                                size_t nmemb, size_t size,
+                                int ((*compar)(const void *, const void *)) :
+                                  itype(_Ptr<int(_Ptr<const T>, _Ptr<const T>)>)
+                                ) : itype(_Ptr<T>);
 
 // TODO: compar needs to have an itype that has bounds
 // on parameters based on size.  Currently we are requiring that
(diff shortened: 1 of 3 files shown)


Prabhuk:
This PR is ready for review.

include/stdlib_checked.h
```diff
@@ -78,11 +78,11 @@ int system(const char *s : itype(_Nt_array_ptr<const char>));
...
 // in the parameter list.
-void *bsearch(const void *key : byte_count(size),
-              const void *base : byte_count(nmemb * size),
+_Itype_for_any(T) void *bsearch(const void *key : itype(_Array_ptr<const T>) byte_count(size),
+              const void *base : itype(_Array_ptr<const T>) byte_count(nmemb * size),
               size_t nmemb, size_t size,
               int ((*compar)(const void *, const void *)) :
-                itype(_Ptr<int(_Ptr<const void>, _Ptr<const void>)>)) :
+                itype(_Ptr<int(_Ptr<const T>, _Ptr<const T>)>)) :
                 byte_count(size);
```
AnnaKornfeldSimpson: I think bsearch should have a return itype of _Ptr<T> since it is picking one element of an array. Ptr can't have byte_count since the type has that information. Similarly I think key should be itype(_Ptr<const T>) and no byte_count.

include/stdlib_checked.h
```diff
@@ -65,7 +65,7 @@ void *aligned_alloc(size_t alignment, size_t size) : byte_count(size);
 _Itype_for_any(T) void *calloc(size_t nmemb, size_t size) : itype(_Array_ptr<T>) byte_count(nmemb * size);
 _Itype_for_any(T) void free(void *pointer : itype(_Array_ptr<T>) byte_count(0));
 _Itype_for_any(T) void *malloc(size_t size) : itype(_Array_ptr<T>) byte_count(size);
-void *realloc(void *pointer : byte_count(1), size_t size) : byte_count(size);
+_Itype_for_any(T) void *realloc(void *pointer : itype(_Array_ptr<T>) byte_count(1), size_t size) : itype(_Array_ptr<T>) byte_count(size);
```
AnnaKornfeldSimpson: We discussed whether the pointer and the return type should match: the only usage of realloc I've seen is to resize an array, so the type does not change.

include/stdlib_checked.h
```diff
@@ -78,12 +78,12 @@ int system(const char *s : itype(_Nt_array_ptr<const char>));
...
-              size_t nmemb, size_t size,
-              int ((*compar)(const void *, const void *)) :
-                itype(_Ptr<int(_Ptr<const void>, _Ptr<const void>)>)) :
-                byte_count(size);
+_Itype_for_any(T) void *bsearch(const void *key : itype(_Ptr<const T>),
+                                const void *base : itype(_Array_ptr<const T>) byte_count(nmemb * size),
+                                size_t nmemb, size_t size,
+                                int ((*compar)(const void *, const void *)) :
+                                  itype(_Ptr<int(_Ptr<const T>, _Ptr<const T>)>)
+                                ) : itype(_Ptr<T>);
```
dtarditi: The return pointer is either null or a pointer to an array element.    A return bounds that would be useful in more contexts would be bounds(base, (Array_ptr<T>) ((char *) base + nmemb * size)).   This provides the additional information that the return value (if it is non-null) is within range of the input array.

AnnaKornfeldSimpson: Interesting! I think what you are saying is that there are contexts where the return value from bsearch would be used as if it were an array_ptr value (e.g. be incremented or decremented)? This is a neat example of determining what a return value for a polymorphic generic function should be - perhaps useful to add it (and explanation) as a case study to the spec?

Prabhuk: Work item created for providing bounds for the return pointer #560

PR #318, opened 2018-09-07
Test subscripting/dereferencing operations on literals

dtarditi:
Add runtime tests of subscripting/dereferencing string literals and compound literals.  Examples of that are code like `"abcdef"[index]`, `*("abcdef" + index)`.   These are one of the last few cases where we weren't inserting runtime bounds checking required by the Checked C semantics.

This matches the Checked C clang compiler PR https://github.com/Microsoft/checkedc-clang/pull/561.

Most of the changes are for adding arguments to test invocations for the new cases.

Fix an error where tests that check dereferencing the result of pointer arithmetic weren't actually checking that.  We needed to add -DPOINTER_ARITHMETIC to the command line.    We have tests that come two variations: checking subscripting, or checking subscripting expanded into a dereference operation and pointer arithmetic.

### DIFF
--- a/include/CMakeLists.txt
+++ b/include/CMakeLists.txt
@@ -19,6 +19,7 @@ set(files
   stdio_checked.h
   stdlib_checked.h
   string_checked.h
+  sys/socket_checked.h
   threads_checked.h
   time_checked.h
   unistd_checked.h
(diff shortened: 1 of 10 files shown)


dtarditi:
Thank you for the review.  The temporaries are not documented yet.  I'm working on updating the spec.

PR #333, opened 2018-11-13
Test generic type arguments at calls.

dtarditi:
Add tests of generic type arguments at calls to generic functions.  
- Test type arguments that should result in success of type checking and type checking errors.
- Test type arguments in scopes with different kinds of checking.
- Test omitting type arguments.  This is allowed in unchecked scopes, but not allowed in checked scopes.  When type arguments are omitted, void is supplied as the type for each type argument instead.

This completes the testing needed for https://github.com/Microsoft/checkedc-clang/issues/572

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,14 @@ and test code.
 - The PDF of the specification is available [here](https://github.com/Microsoft/checkedc/releases).
 - Compilers are available [here](https://github.com/Microsoft/checkedc-clang/releases).
 
+# Papers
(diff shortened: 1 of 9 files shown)


dtarditi:
Add tests of generic type arguments at calls to generic functions.  
- Test type arguments that should result in success of type checking and type checking errors.
- Test type arguments in scopes with different kinds of checking.
- Test omitting type arguments.  This is allowed in unchecked scopes, but not allowed in checked scopes.  When type arguments are omitted, void is supplied as the type for each type argument instead.

This completes the testing needed for https://github.com/Microsoft/checkedc-clang/issues/572

PR #350, opened 2019-05-07
Update README.md

oalign:
removed job proposal it is not relevant anymore


dtarditi:
Thanks for noticing the issue with the URL.  The position description was being reposted and the URL was only temporarily unavailable.

PR #361, opened 2019-06-18
Add support for ARM lit testing #618

mgrang:


mgrang:
This is now no longer needed. Will close this once #372 lands.

PR #370, opened 2019-07-19
Tests for generic structs

abeln:
This PR adds multiple tests that exercise the functionality in https://github.com/microsoft/checkedc-clang/pull/638

### DIFF
--- a/tests/typechecking/generic_structs.c
+++ b/tests/typechecking/generic_structs.c
@@ -5,7 +5,7 @@
 //
 // Test that only the right type variables are in scope.
 //
-struct Foo _For_any(T) {
+struct T1_Foo _For_any(T) {
   T *x;
   U *y; // expected-error {{unknown type name 'U'}}
 };
@@ -15,7 +15,7 @@ struct Foo _For_any(T) {
 // a struct definition (because the size of the underlying type
 // is unknown).
 //
-struct Bar _For_any(T) {
+struct T2_Foo _For_any(T) {
   T *x;
   T  y; // expected-error {{field has incomplete type 'T'}}
 };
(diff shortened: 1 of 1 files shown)


tests/typechecking/generic_structs.c
```diff
@@ -0,0 +1,267 @@
+// Test type checking of generic structs.
+//
+// RUN: %clang_cc1 -verify -verify-ignore-unexpected=note %s
+
+//
+// Test that only the right type variables are in scope.
+//
+void TestParamScope() {
```
sunnychatterjee: Another option for you would be to mark the test properties as namespaces rather than functions. Of course, you should include functions inside these namespaces wherever they make sense. 

For example:

namespace TestParamScope
{
struct Foo _For_any(T) {
   T *x;
   U *y; // expected-error {{unknown type name 'U'}}
  };
}

Similarly, for the remaining tests below.

abeln: The tests are C code, so we can't use namespaces.

abeln:
> Thanks for these tests. Overall, they look good. For recursive cycles, I think you should try to do some additional types that could trigger a cycle (such as function types and array types. 

Added tests for functions and array types.

Could you also add some tests that involve bounds declarations and generic structures? It would be helpful to have some uses of type variables in the bounds expressions also. I'm not looking for thorough testing of the interaction of generic structures and bounds declaration: I just want to make sure that substitution works properly for bounds expressions and that the compiler doesn't crash.

Can I punt on adding theses until we get itype-forany working for generic structs? I have those changes in a separate branch https://github.com/abeln/checkedc-clang/commits/itype-forany and would like to handle them in a separate PR.

abeln:
@dtarditi PTAL

PR #374, opened 2019-07-31
Enable dynamic_checking tests for ARM

mgrang:
Added the following to all tests under dynamic_checking:
1.  %checkedc_target_flags to all clang compile RUN lines.
2. %checkedc_rununder to all execution RUN lines.

### DIFF
--- a/tests/dynamic_checking/bounds/array-bounds-decls.c
+++ b/tests/dynamic_checking/bounds/array-bounds-decls.c
@@ -3,20 +3,20 @@
 // override bounds based on the size of the 1st 
 // dimension of the array.
 //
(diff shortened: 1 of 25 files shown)


mgrang:
I noticed that some of the checkedc tests have RUN lines without the colon like `RUN %t` instead of `RUN: %t`. Such lines are not actually run. If I add back the colon then I see that the test runs and fails.
For e.g.: https://github.com/microsoft/checkedc/blob/master/tests/dynamic_checking/bounds/subscript_call_expr_opt.c

David says this is an error. I will let this patch merge and fix this in a later patch.

PR #377, opened 2019-08-15
Added tests for the "*" operator of _MMSafe_ptr and added customized memory allocator & deallocator.

jzhou76:
Tests include:
- using local variable mmsafe_ptr to read and write struct fields
- mmsafe_ptr assignment
- reads and writes through the new mmsafe_ptr after pointer assignment
- passing mmsafe_ptr to functions
- using parameter mmsafe_ptr to read and write struct fields
- use-after-free testing

Also added a customized memory allocator and a deallocator. See the
code for more details.

### DIFF
--- a/tests/safe-memory-management/dereference.c
+++ b/tests/safe-memory-management/dereference.c
@@ -1,0 +1,140 @@
+// RUN: %clang %s
+// expected-no-diagnostics
+//
+
+#include <stdio.h>
+#include <stdbool.h>
+#include "safe_mm_checked.h"  
+
+typedef struct data {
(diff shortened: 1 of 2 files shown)


jzhou76:
> Looks good.
> 
> There are two things that you could do next to improve the test. This can be done in a future commit:
> 
> 1. Have the test take an argument indicating what function to run.
> 2. Add a signal handler to catch the signal and have the program exit gracefully.  The other runtime tests show how to do this.

Thanks for the suggestions. Can you point a sample test file to me? Especially for the first one.

PR #378, opened 2019-08-19
Add runtime tests for null ptr arithmetic

mgrang:


mgrang:
This depends on https://github.com/microsoft/checkedc-clang/pull/663

mgrang:
> we should add additional tests of the different kinds of variables. The tests currently only test global variables. In my experience with clang, clang can go down different paths for local variables and parameters, so it is important to test them.

Done https://github.com/microsoft/checkedc-clang/issues/670

PR #383, opened 2019-09-20
Add local variable and parameter null ptr arithmetic tests

kkjeer:
Add runtime null pointer arithmetic tests to test local variables and parameters, in addition to the existing tests for global variables (see issue [#670](https://github.com/microsoft/checkedc-clang/issues/670)).


mgrang:
While you are at it can you also rename the test file to `ptr-arith.c`?

PR #393, opened 2019-11-26
Matching tests to the new logic on handling empty/invalid ranges

saeednj:
This is to update the tests to match the new handling of empty and invalid ranges. The corresponding code for the new handling is in this [pull-request](https://github.com/microsoft/checkedc-clang/pull/707).

### DIFF
--- a/tests/dynamic_checking/bounds/deref.c
+++ b/tests/dynamic_checking/bounds/deref.c
@@ -7,7 +7,7 @@
 //
 // The following lines are for the clang automated test suite.
 //
-// RUN: %clang %s -o %t1 -DTEST_READ -Werror -Wno-unused-value -Wno-check-memory-accesses -Wno-check-bounds-decls-unchecked-scope %checkedc_target_flags
+// RUN: %clang %s -o %t1 -DTEST_READ -Werror -Wno-unused-value -Wno-check-bounds-decls-unchecked-scope %checkedc_target_flags
 // RUN: %checkedc_rununder %t1 pass1 | FileCheck %s --check-prefixes=CHECK,CHECK-PASS,PASS-1-READ
 // RUN: %checkedc_rununder %t1 pass2 | FileCheck %s --check-prefixes=CHECK,CHECK-PASS,PASS-2-READ
 // RUN: %checkedc_rununder %t1 pass3 | FileCheck %s --check-prefixes=CHECK,CHECK-PASS,PASS-3-READ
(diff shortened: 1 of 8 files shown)


tests/static_checking/bounds_decl_checking.c
```diff
@@ -333,7 +333,7 @@ extern void check_exprs_nullterm(nt_array_ptr<int> arg1 : bounds(unknown),
 
   // assignment through pointer
   *t1 = 1;            // expected-error {{expression has unknown bounds}}
-  *t2 = 2;            // expected-warning {{out-of-bounds memory access}}
```
dtarditi: Could you file a GitHub issue for this?  We should give an error when assign a non-zero value to the upper bound of an nt_array_ptr.

tests/dynamic_checking/bounds/deref.c
```diff
@@ -7,7 +7,7 @@
 //
 // The following lines are for the clang automated test suite.
 //
-// RUN: %clang %s -o %t1 -DTEST_READ -Werror -Wno-unused-value -Wno-check-memory-accesses %checkedc_target_flags
+// RUN: %clang %s -o %t1 -DTEST_READ -Werror -Wno-unused-value -Wno-check-memory-accesses -Wno-check-bounds-decls-unchecked-scope %checkedc_target_flags
```
dtarditi: I think the flag -Wno-check-memory-accesses can be deleted from all of these command lines.   I am surprised you had to add -Wno-check-bounds-decls-unchecked-scope.   Were there specific lines in the file that triggered this warning?

tests/dynamic_checking/bounds/deref.c
```diff
@@ -208,7 +210,8 @@ void passing_test_3(array_ptr<int> a : count(len), int len) {
 // Bounds describe empty range, no deref
 void failing_test_1(void) {
   int a checked[2] = { 0, 0 };
-  array_ptr<int> b : bounds(a, a) = a;
+  int k = 0;
```
dtarditi: The compiler will get smart enough soon to defeat this - i.e., it would understand when a variable value is constant within a within a function.  It might be better to make this a parameter.

tests/dynamic_checking/bounds/deref_opt.c
```diff
@@ -13,7 +13,7 @@
 //
 // The following lines are for the clang automated test suite.
 //
-// RUN: %clang %S/deref.c -o %t1 -DTEST_READ -Werror -Wno-unused-value  -Wno-check-memory-accesses -O3 %checkedc_target_flags
+// RUN: %clang %S/deref.c -o %t1 -DTEST_READ -Werror -Wno-unused-value  -Wno-check-memory-accesses -Wno-check-bounds-decls-unchecked-scope -O3 %checkedc_target_flags
```
dtarditi: The comments for deref.c apply to this file too.

tests/dynamic_checking/bounds/deref.c
saeednj: Yes, it seems that -Wno-check-memory-accesses has no effect.
The -Wno-check-bounds-decls-unchecked-scope was needed because in cases that we see an invalid range declaration, we give a warning and this flag ignores that warning. For example, for the declaration at line 226 of deref.c

tests/dynamic_checking/bounds/deref.c
saeednj: Sure! I made function arguments and moved the constants to function calls.

tests/dynamic_checking/bounds/deref.c
saeednj: Yes, -Wno-check-memory-accesses was for out-of-bounds memory access warnings.
I removed them.

PR #395, opened 2020-02-11
Introduce read/write sets for bounds declaration checking description.

dtarditi:
This changes simplifies the description of bounds declaration checking to use read/write sets of pending variable sets.  

Currently we are using a notion of pending variables in the description of checking of bounds declarations. They are computed by another function that is called during checking. This doesn't extend well to lvalue expressions because lvalue expressions depend on variables, which may be modified within an expression.

This change replaces pending variables with read/write summaries.  This allows a bottom computation of a summary of the effects of an expression.  We can then determine if the read/write summary for expression with an unspecified order of evaluation with respect to each other conflict. The read/write summaries for lvalue expressions can use variables defined in known frames of reference (states after an expression is evaluated).

There are some examples that still need to be updated.    I plan to add text to this change that notes where examples still need to be updated.

### DIFF
--- a/spec/bounds_safety/checking-variable-bounds.tex
+++ b/spec/bounds_safety/checking-variable-bounds.tex
@@ -1194,7 +1194,7 @@ Let $B = bounds(v, v + n)$ and $G =\{ v \}$.
 \item Let $RB = $ \boundsunknown (Checked C does not allow bounds for values stored in arrays).
 \end{enumerate}
 \end{enumerate}
-\item If $e$ has the form $*e1$,  let $(B, \mathit{UC}, EQ, \_) = Check(e1, C, EQ, No)$.  Let $G =\emptyset$ and $RB = $ \boundsunknown (Checked C does not allow bounds for values pointed to by pointers).
+\item If $e$ has the form $*e1$,  let $(B, \mathit{UC}, EQ, \_, R, W) = Check(e1, C, EQ, No)$.  Let $G =\emptyset$ and $RB = $ \boundsunknown (Checked C does not allow bounds for values pointed to by pointers).
 \item If $e$ has the form $e1[e2]$,
 \label{list:check-unordered-subscript-operands}
 \begin{enumerate}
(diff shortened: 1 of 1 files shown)


spec/bounds_safety/checking-variable-bounds.tex
```diff
@@ -1208,18 +1194,21 @@ \subsection{Checking expressions with unordered evaluation of operands}
 \item Let $RB = $ \boundsunknown (Checked C does not allow bounds for values stored in arrays).
 \end{enumerate}
 \end{enumerate}
-\item If $e$ has the form $*e1$,  let $(B, \mathit{UC}, EQ, \_) = Check(e1, C, EQ, PV, No)$.  Let $G =\emptyset$ and $RB = $ \boundsunknown (Checked C does not allow bounds for values pointed to by pointers).
+\item If $e$ has the form $*e1$,  let $(B, \mathit{UC}, EQ, \_, R, W) = Check(e1, C, EQ, No)$.  Let $G =\emptyset$ and $RB = $ \boundsunknown (Checked C does not allow bounds for values pointed to by pointers).
```
kkjeer: Is `let (B, UC, EQ, _, R, W) = Check(...)` supposed to be `let (B, UC, UEQ, _, R, W) = Check(...)`? It seems like UEQ for `*e1` should be equal to UEQ for `e1`.

spec/bounds_safety/checking-variable-bounds.tex
```diff
@@ -1326,13 +1314,19 @@ \subsection{Control-flow expressions}
 to produce the set of sets.  Empty or singleton sets are removed.
 \item Let $G = G_1 \cap G_2$.  If $e$ has no side-effects,add $\{ e \}$ to $G$.  If a
 temporary was introduced for the value of $e$,  also add the temporary to $G$.
+\item Let $R = RC \cup R_1 \ cup R_2$.  Let $W = WC \cup W_1 \cup W_2$.
```
kkjeer: typo: `\ cup` => `\cup`

spec/bounds_safety/checking-variable-bounds.tex
```diff
@@ -1326,13 +1314,19 @@ \subsection{Control-flow expressions}
...
+\item Make sure that all assignments to variables used in $DC^\prime$ have been seen before
+validating $DC^prime$ (otherwise the validation may be incorrect).
```
kkjeer: typo: `DC^prime` => `DC^\prime`

dtarditi:
@kkjeer, thanks for pointing out these typos.  I found some additional ones as well and fixed them.  This is ready for further review.

PR #398, opened 2020-03-31
Add tests for bounds-safe interface assignments

kkjeer:
This PR adds Checked C tests for clang PR [#813](https://github.com/microsoft/checkedc-clang/pull/813), which fills in two missing cases for bounds-safe interface assignment compatibility.

### DIFF
--- a/tests/typechecking/interop.c
+++ b/tests/typechecking/interop.c
@@ -607,3 +607,23 @@ void g82(callback_fn3 fn) {
   int arr checked[10][10];
   (*fn)(arr);
 }
+
+//
+// Test assignments involving non-variable lvalues
+// and bounds-safe interfaces
+//
+
+// Assign a checked pointer to the dereference of an
+// unchecked pointer with a bounds-safe interface.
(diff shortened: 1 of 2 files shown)


tests/typechecking/interop.c
```diff
@@ -607,3 +607,23 @@ void g82(callback_fn3 fn) {
...
+//
+// Test assignments involving non-variable lvalues
+// and bounds-safe interfaces
+//
+
+// Assign a checked pointer to the dereference of an
+// unchecked pointer with a bounds-safe interface.
+void g90(int **interop_ptr : itype(ptr<ptr<int>>), ptr<int> checked_ptr) {
+  *interop_ptr = checked_ptr;
+  *(interop_ptr - 1) = checked_ptr;
```
dtarditi, line 620: Pointer arithmetic shouldn't really be done on a variable with a declared interface type of `ptr`.  There's a semantic inconsistency in the code.  I suggest you change this to  have an `itype(_array_ptr<..`.  You'll need to also declare a bounds for the `interop_ptr`.

Issue #397, opened 2020-03-20
Update spec to include missing cases for bounds-safe interface assignment compatibility


kkjeer:
The current rules for assignment compatibility of expressions involving bounds-safe interfaces are missing two cases.

The first case is related to assignments involving nested pointers.
An assignment `*p = e` can pass type checking if:
* `p` has an unchecked pointer type and a bounds-safe interface, and
* `e` has a checked pointer type, and
* the referent type of `e` is assignment compatible with the referent type of the referent type of `p`

The second case is related to assignments of an unchecked pointer with an `_Nt_array_ptr` bounds-safe interface to an `_Nt_array_ptr` variable.
An assignment `v = e` can pass type checking if:
* `v` has type `_Nt_array_ptr<T>`, and
* `e` has unchecked pointer type and a bounds-safe interface of `_Nt_array_ptr<U>`, and
* the referent types `T` and `U` are assignment compatible.

PR #399, opened 2020-04-01
Update spec to include missing cases for bounds-safe interface assignments

kkjeer:
See issue #397 

This PR adds some missing cases for bounds-safe interface assignment compatibility to section 6.3.6 (Interoperation - Type checking).

### DIFF
--- a/spec/bounds_safety/interoperation.tex
+++ b/spec/bounds_safety/interoperation.tex
@@ -954,6 +954,18 @@ They are done at:
      implicitly to the unchecked pointer type.
 \item
     Member assignments: a similar conversion is done for member assignments.
+\item Assignments to pointer dereferences: if the pointer dereference expression
+  being assigned to has an unchecked pointer type and a bounds-safe interface, the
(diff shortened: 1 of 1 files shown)


spec/bounds_safety/interoperation.tex
```diff
@@ -954,6 +954,24 @@ \subsection{Type checking}
      implicitly to the unchecked pointer type.
 \item
     Member assignments: a similar conversion is done for member assignments.
+\item Assignments to pointer dereferences: if the pointer dereference expression
```
kkjeer, line 957: I added full text for the assignments to pointer deferences and array subscripts to be explicit, but these items could be more similar to the item for member assignments, e.g. "Pointer dereference assignments: a similar conversion is done for pointer deference assignments".

dtarditi: This look good to me.  Does the spec need to define what we mean by an (lvalue) expression having a bounds-safe interface?

PR #400, opened 2020-05-11
Fix checkedc unit test after bounds widening

mgrang:
After we have established that the value at the upper bound of an nt_array_ptr
s is non-null, we widen its bounds by 1. So the new upper bound of s is
(curr upper bound + 1). So in order to test writing at the upper we need to
write to *(s+1) instead of *s.


mgrang:
After the bounds widening context update PR https://github.com/microsoft/checkedc-clang/pull/821 merges this  will fix the failure seen in https://dev.azure.com/msresearch/CheckedC/_build/results?buildId=486239&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b&t=f2cd6026-4e71-5279-c72c-79cca5121781

PR #402, opened 2020-05-15
Bounds context: update after assignments (tests)

kkjeer:
This PR updates checked-c tests to add expected warnings and errors since the checkedc-clang PR [836](https://github.com/microsoft/checkedc-clang/pull/836) updates the observed bounds context after an assignment to a variable.

### DIFF
--- a/tests/parsing/pointer_bounds_cast.c
+++ b/tests/parsing/pointer_bounds_cast.c
@@ -97,7 +97,7 @@ extern void f10() {
 extern void f11() {
   array_ptr<int> r : count(3) = 0;
   ptr<int> q = 0;
-  r = _Assume_bounds_cast<array_ptr<int>>(h4(), bounds(r, r + 4)  rel_align(int));
+  r = _Assume_bounds_cast<array_ptr<int>>(h4(), bounds(r, r + 4)  rel_align(int)); // expected-error {{expression has unknown bounds}}
   q = _Assume_bounds_cast<ptr<int>>(h4());
 }
 
--- a/tests/parsing/pointer_bounds_cast.c
+++ b/tests/parsing/pointer_bounds_cast.c
@@ -97,7 +97,16 @@ extern void f10() {
 extern void f11() {
   array_ptr<int> r : count(3) = 0;
   ptr<int> q = 0;
+
+  // The declared bounds of h4() use the value of r, but r is overwritten
+  // in the assignment. The value of r (used in the declared bounds (r, r + 4))
+  // is lost, so the inferred bounds for the cast expression are unknown.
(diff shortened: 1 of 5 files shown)


tests/static_checking/bounds_decl_checking.c
```diff
@@ -33,7 +33,7 @@ extern void check_exprs(int *arg1, ptr<int> arg2, array_ptr<int> arg3,
   // address-of
   int tmp1 = 0;
   arg4 = &tmp1;
-  arg4 = &*arg4;
+  arg4 = &*arg4;          // expected-error {{expression has unknown bounds}}
```
kkjeer, line 36: This could be confusing to the user to get an error message here, since this is effectively the same as the assignment `arg4 = arg4`. In the assignment `arg4 = arg4`, the original value of `arg4` is `arg4`, so there is no error. However, in this assignment `arg4 = &*arg4`, `arg4` has no original value, so the observed bounds for the source of the assignment are `bounds(unknown)`.

A potential fix would be to modify the IsInvertible and Inverse methods so that the original value of `&*p` with respect to the variable `p` is `p`. Then, an assignment `p = &*p` would have the same behavior as the assignment `p = p`. @dtarditi would this be the correct behavior?

kkjeer: This issue will be addressed once the checkedc-clang PR [839](https://github.com/microsoft/checkedc-clang/pull/839) is merged. `&*arg4` will have an inverse of `arg4`.

kkjeer: This is now addressed after merging checkedc-clang [839](https://github.com/microsoft/checkedc-clang/pull/839).

tests/parsing/pointer_bounds_cast.c
```diff
@@ -97,7 +97,7 @@ extern void f10() {
 extern void f11() {
   array_ptr<int> r : count(3) = 0;
   ptr<int> q = 0;
-  r = _Assume_bounds_cast<array_ptr<int>>(h4(), bounds(r, r + 4)  rel_align(int));
+  r = _Assume_bounds_cast<array_ptr<int>>(h4(), bounds(r, r + 4)  rel_align(int)); // expected-error {{expression has unknown bounds}}
```
dtarditi, line 100: It is worth adding a comment explaining what is broken here.  We declare the result of h4 to have bounds based on r, but overwrite r, invaliding the bounds of the entire _Assume_bounds_cast expression (since r has now changed).

Could you add an example of how to property initialize `r`?   Does
```
r = _Assume_bounds_cast<array_ptr<int>>(h4(), count(4) )
```
work?

kkjeer: `r = _Assume_bounds_cast<array_ptr<int>>(h4(), count(4))`. I've added it to the test.

PR #412, opened 2020-08-03
Update tests to reflect improved bounds checking error messages

dopelsunce:
Update tests for the change https://github.com/microsoft/checkedc-clang/pull/893.

### DIFF
--- a/tests/parsing/pointer_bounds_cast.c
+++ b/tests/parsing/pointer_bounds_cast.c
@@ -14,7 +14,7 @@ extern void f1() {
 extern void f2() {
   char p[10];
   array_ptr<int> a : count(1) = 0;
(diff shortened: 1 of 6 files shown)


dopelsunce:
Yes, all updates from checkedc-clang/893 passed this test. 

Actually, there is no test case for "bounds of a variable are unknown after initialization" in this `checkedc/tests` folder. But there is cases for "bounds of a variable are invalid after initialization".

kkjeer:
> Yes, all updates from checkedc-clang#893 passed this test (manually).
> 
> Actually, there is no test case for "bounds of a variable are unknown after initialization" in this `checkedc/tests` folder. But there are cases for "bounds of a variable are invalid after initialization".

Thanks for clarifying. Can you add a test to one of these files (maybe bounds_decl_checking.c) for "bounds of a variable are unknown after initialization"?

PR #419, opened 2020-09-22
spec: modf returns a double.

mattmccutchen-cci:


mgrang (approved):
Thanks for fixing this! LGTM.

PR #420, opened 2020-09-22
spec: Fix typo that blocks the build when parskip.sty is not installed.

mattmccutchen-cci:
I'll probably install `parskip.sty` to match what your configuration must have been to successfully produce the released specs, but as long as you have this code path at all, it should work.


dtarditi (approved):
Thanks!

PR #422, opened 2020-10-20
Update tests for conditional rvalue bounds

kkjeer:
This PR adds expected warnings and errors to tests that result from inferring rvalue bounds for conditional expressions (in [checkedc-clang/924](https://github.com/microsoft/checkedc-clang/pull/924)).

### DIFF
--- a/tests/typechecking/checked_arrays.c
+++ b/tests/typechecking/checked_arrays.c
@@ -514,7 +514,7 @@ extern void check_condexpr(int val) {
   array_ptr<int> t5 = val ? r : 0;
   array_ptr<int> t5a = val ? v : 0;
   array_ptr<int> t6 = val ? 0 : r;
-  nt_array_ptr<int> t6a = val ? 0 : v;
+  nt_array_ptr<int> t6a = val ? 0 : v; // expected-warning {{cannot prove declared bounds for 't6a' are valid after initialization}}
   array_ptr<float> t7 = val ? u : 0;
   array_ptr<float> t8 = val ? 0 : u;
 }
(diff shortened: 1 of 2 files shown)


tests/typechecking/checked_arrays.c
```diff
@@ -514,7 +514,7 @@ extern void check_condexpr(int val) {
   array_ptr<int> t5 = val ? r : 0;
   array_ptr<int> t5a = val ? v : 0;
   array_ptr<int> t6 = val ? 0 : r;
-  nt_array_ptr<int> t6a = val ? 0 : v;
+  nt_array_ptr<int> t6a = val ? 0 : v; // expected-warning {{cannot prove declared bounds for 't6a' are valid after initialization}}
```
dtarditi, line 517: I am surprised by this.  I would have expected 0 to have bounds(any), the bounds for v to be count(5), and the overall bounds for the two arms to be count(5).    What are the actual bounds the compiler is inferring for the two arms?

dtarditi (commented):
I am confused why these changes are necessary.  It seems like something unexpected is happening to bounds for conditional expressions that evaluate to an `nt_array_ptr` and involve implicit casts for the arms.

tests/typechecking/checked_arrays.c
kkjeer: The compiler infers bounds(any) for 0, bounds(v, v + 5) for v, and bounds(v, v + 5) as the overall bounds for the two arms. v and t6a are not known to be equal, so the compiler cannot prove that bounds(v, v + 5) imply bounds(t6a, t6a + 0).

kkjeer: Created [checkedc-clang/934](https://github.com/microsoft/checkedc-clang/issues/934) to track this follow-up work.

mgrang (approved):
LGTM, unless David has more comments.

tests/typechecking/checked_arrays.c
dtarditi: Thanks for the explanation and opening the issue to track this.  I agree that this should be handled as a separate follow-up issue.

dtarditi (approved):
Looks good, thanks.

PR #432, opened 2021-01-06
Assorted typo and other minor fixes to the specification.

mattmccutchen-cci:
I'm just giving you notice that this is coming.  I'm guessing you'd rather wait until my batch of fixes is complete before going to the overhead of reviewing it.

### DIFF
--- a/spec/bounds_safety/introduction.tex
+++ b/spec/bounds_safety/introduction.tex
@@ -306,9 +306,9 @@ Two specific design principles are adopted based on these principles:
   inventing new notations. Many systems are hybrid C/C++ systems, so
   this approach fits with the principle of clarity. It also enables
   incremental adoption. One of the design goals of C++ has been that C
-  is a subset of C++.  It is a design goal to allow Checked C to be a
-  subset of C++ too.\feedback{Matt McCutchen}{ISTM reality must be more
-    complicated than this, given that Checked C has syntax extensions that are not in C++.}
+  is a subset of C++.  It is a design goal to allow Checked C extensions
+  to be amenable to being added to C++ too, so that Checked C is a subset
+  of a Checked C++.
 \end{enumerate}
 
 \section{Notation}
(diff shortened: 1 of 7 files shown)


mattmccutchen-cci:
I think this is all that is coming from me for now.  Today I read (or skimmed) the parts of the specification I thought would be important to me as a 3C developer, and I decided to fix typos as I went.

spec/bounds_safety/introduction.tex
```diff
@@ -302,12 +302,13 @@ \section{Principles for extensions}
...
 \item
-  Adopt existing notations from C++ when it meets our needs, instead of
+  Adopt existing notations from C++ when they meet our needs, instead of
   inventing new notations. Many systems are hybrid C/C++ systems, so
   this approach fits with the principle of clarity. It also enables
   incremental adoption. One of the design goals of C++ has been that C
   is a subset of C++.  It is a design goal to allow Checked C to be a
-  subset of C++ too.
+  subset of C++ too.\feedback{Matt McCutchen}{ISTM reality must be more
+    complicated than this, given that Checked C has syntax extensions that are not in C++.}
```
mattmccutchen-cci, line 311: What should we say here about the relationship between Checked C and C++?  (Starting an unresolved thread to help ensure this TODO comment isn't accidentally merged.)

dtarditi: Hmm, I see how this sentence could be misinterpreted.   I think it is fine to be more direct: It is a design goal to allow Checked C extensions to be amenable to being added to C++ too, so that Checked C is a subset of a Checked C++.

dtarditi (commented):
All of these minor fixes look good to me.   I think these changes are fine provided that the comment about the relationship of Checked C and C++ is updated.

spec/bounds_safety/introduction.tex
mattmccutchen-cci: I took your text. I can think of other ways to word this that might be better, but I don't think it's worth the time to nitpick.

mattmccutchen-cci:
Thanks @dtarditi for the review.  I hope this is OK now.

mattmccutchen-cci:
@dtarditi Ping.  I imagine you're busy, but if it will take you only a minute or two to verify that I addressed your concern, perhaps it makes sense to merge this rather than let it sit indefinitely?

dtarditi (commented):
Looks good, thanks for addressing my feedback!

PR #433, opened 2021-01-26
Fix failing test resulting from free variables merge

kkjeer:
The parsing/pointer_bounds_cast.c test expected an error where the bounds of 'x' are invalid after a dynamic bounds cast. The actual error is that it is impossible to prove that the bounds of 'x' are valid since there is no relational information between the variables 'cache1' in the inferred bounds and 'x' in the declared bounds.


dtarditi (commented):
LGTM.

sulekhark (commented):
LGTM.

PR #434, opened 2021-01-29
Fixed for MacOS

aaronjeline:
Mac OS sets macros for string functions and has `_FORTIFY_SOURCE` set to `2`.

### DIFF
--- a/include/string_checked.h
+++ b/include/string_checked.h
@@ -60,28 +60,28 @@
 // - Fortifying,     No Macro:     we need the definition
 // - Not Fortifying, No Macro:     we need the definition
 
-#if _FORTIFY_SOURCE == 0 || !defined(memcpy)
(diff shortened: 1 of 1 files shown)


sulekhark:
Hi @aaronjeline,  thank you for your work on this and for submitting this PR. It will be great if the following review comments are addressed:
1. The comments on lines 54 thro' 61 need to be updated stating why the definitions are necessary in the macOS environment in spite of FORTIFY_SOURCE being greater than 0.
2. The fix should be applicable only in the macOS environment when FORTIFY_SOURCE == 2.  It should not change definitions in non-macOS environments when FORTIFY_SOURCE ==2.
3. The same fix is also applicable to some definitions in stdio_checked.h.

dtarditi:
@aaronjeline, thanks for the changes!  Could you address the pull request feedback so that we can commit this?

dtarditi:
These changes have already been brought into the Checked C repo, so closing it.

PR #444, opened 2021-04-27
Update documentation for using function pointers

2000jedi:
* Include documentation on function pointer types and type annotations.
* Related to issue [#1027](https://github.com/microsoft/checkedc-clang/issues/1027) in the clang repo.

### DIFF
--- a/spec/bounds_safety/interoperation.tex
+++ b/spec/bounds_safety/interoperation.tex
@@ -862,6 +862,28 @@ void f(int **p : itype(array_ptr<int> checked[10]));
 void f(int **p : itype(int *checked[20]));
 \end{lstlisting}
 
+\subsection{Function pointer type as parameter}
+Function pointers that are parameters in a bounds-safe interface specification
+can themselves have a bounds-safe interface type.  This allows both checked and
+unchecked function pointers to be passed as actual arguments.
+Here is an example:
+\begin{lstlisting}
+int apply_fun (
+  char *buffer : itype(array_ptr<char>) count(length),
+  size_t length,
+  int ((*fun)(char *buf, size_t len)) :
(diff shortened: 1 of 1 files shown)


spec/bounds_safety/interoperation.tex
```diff
@@ -885,6 +885,18 @@ \subsection{Bounds expression and type annotations}
 an interface type and a bounds expression are both specified, the interface type must
 be an \arrayptr\ type or checked array type.
 
+\subsection{Function pointer types and type annotations}
+Function pointers types are compatible with bounds-safe interfaces.  It requires
+an extra pair of parenthesis around the function pointer declaration before the type annotation.
+This allows a function pointer to be called using both checked and unchecked
+pointers as actual arguments.  For example, in the function pointer declaration
```
sulekhark, line 892: This description could be made more clear as suggested below:

Function pointers that are parameters in a bounds-safe interface specification can themselves have a bounds-safe interface type. This allows both checked and unchecked function pointers to be passed as actual arguments. Here is an example:
```
int apply_fun (char *buffer : itype(arry_ptr<char>),
               int ((*clear)(char *buf, size_t len)) :
               itype(ptr<int(array_ptr<char> buf : count(len), size_t len)>)
              );
```
In the above example `clear` is a function pointer. Note that an extra pair of parentheses is required around the function pointer declaration before the type annotation. This is necessary to disambiguate parsing. Without the extra pair of parentheses, the compiler attempts to bind the itype `(ptr<int(array_ptr<char> buf : count(len), size_t len)>)` to the return value of the function pointed to by `clear`, which will result in a compilation error.

sulekhark (commented):
Thank you very much for proactively initiating this documentation. Please see the inlined review comments.

sulekhark (approved):
LGTM! Thank you once again for your initiative!

PR #448, opened 2021-05-26
Batch of checked header changes from CCI 2021-05

mattmccutchen-cci:
As previously discussed with Sulekha, CCI is preparing to submit a batch of changes to the checked headers that we made in support of our recent porting efforts.  I'm going ahead and submitting a draft PR so that I can more easily refer to some of our changes in issues that I'm about to file.  Warning: I may still overwrite the commits of the PR via force-push.

One part of the PR that I believe is in final form and that you can go ahead and review if you wish is the change to make `#include`-ing a wrapper header an error if the original header does not exist on the system.  This change is currently in [this commit](https://github.com/microsoft/checkedc/commit/a908b533ac427dff53962840c11367411261abab) (though I may overwrite the commit later).  It might make sense to make this change into a separate PR (at the cost of some additional overhead) because it is different in nature from all our other changes; let me know if you'd like me to do so.

The corresponding draft PR for the tests in the checkedc-clang repository is microsoft/checkedc-clang#1064.

### DIFF
--- a/include/CMakeLists.txt
+++ b/include/CMakeLists.txt
@@ -12,12 +12,22 @@ set(files
   assert.h
   errno_checked.h
   errno.h
+  fcntl.h
+  fcntl_checked.h
   fenv_checked.h
   fenv.h
+  grp_checked.h
+  grp.h
   inttypes_checked.h
   inttypes_checked_internal.h
   math_checked.h
   math.h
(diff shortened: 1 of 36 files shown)


dtarditi:
These changes have already been brought into the Checked C repo.

PR #455, opened 2021-06-04
[LValue Generalization] Update tests for validating member bounds

kkjeer:
This PR updates test files to account for the new compiler behavior introduced by [checkedc-clang/1083](https://github.com/microsoft/checkedc-clang/pull/1083): validating the bounds context for member expressions. The compiler now emits errors and warnings for statements that invalidate member bounds by assigning to member expressions that are used in member bounds.


sulekhark (approved):
LGTM! Thank you!

PR #463, opened 2021-07-22
Get variable and constant: update tests

kkjeer:
This PR updates tests for [checkedc-clang/1135](https://github.com/microsoft/checkedc-clang/pull/1135). When the new method `GetVariableAndConstant` is used to compare two upper bounds expressions, the bounds checker is able to prove that the inferred bounds `bounds(arr, arr + 5)` are equivalent to the required cast bounds `bounds((_Array_ptr<int _Checked[]>)&arr, (_Array_ptr<int _Checked[]>)&arr + 1))`


sulekhark (approved):
LGTM! Thank you!

mgrang (approved):
LGTM

PR #465, opened 2021-08-10
Check return value bounds: update tests

kkjeer:
This PR updates test files to account for the new compiler behavior in [checkedc-clang/1150](https://github.com/microsoft/checkedc-clang/pull/1150): checking that the inferred bounds of a return expression imply the declared bounds (if any) for the enclosing function.


sulekhark (approved):
LGTM!

PR #466, opened 2021-08-18
Treat address-of array subscripts the same way as address-of dereferences: update tests

kkjeer:
This PR updates the tests to reflect the updated type checking behavior for address-of array subscripts in [checkedc-clang/1163](https://github.com/microsoft/checkedc-clang/pull/1163). If an expression `e` has type `T`, then `&e[idx]` and `&idx[e]` also have type `T`.


sulekhark (approved):
LGTM! Thank you!

PR #472, opened 2021-09-01
Bounds checking for pointer dereferences and array subscripts: update test

kkjeer:
This PR updates some expected error messages to reflect the behavior in [checkedc-clang/1176](https://github.com/microsoft/checkedc-clang/pull/1176): for a pointer dereference or array subscript with target bounds, bounds validation uses the existing framework in `UpdateAfterAssignment` and `ValidateBoundsContext` to validate these expressions. This means that the error messages emitted for pointer dereferences or array subscripts whose inferred bounds are unknown have been updated to be consistent with the error messages for checking variables and member expressions, so the expected error messages in the test file bounds_decl_checking.c are updated.


sulekhark (approved):
LGTM! Thank you!

PR #473, opened 2021-09-02
Fix a test that invokes variadic functions in checked scope

mgrang:
https://github.com/microsoft/checkedc-clang/pull/1174 added support to call
variadic functions like printf/scanf, etc within checked scope. As a result,
tests that check for errors if such functions are invoked in checked scope
started failing. We fix one such test in this PR.


sulekhark (approved):
LGTM! Thank you!

kkjeer (approved):
LGTM!

PR #474, opened 2021-09-09
Get lvalue observed bounds: update test

kkjeer:
This PR updates the bounds-safe-interface test file to account for the changes in [checkedc-clang/1183](https://github.com/microsoft/checkedc-clang/pull/1183).

With this change, the bounds for the value of the member expression `s1->f` are the observed bounds for `s1->f` as recorded in `CheckingState.ObservedBounds`. This means that, after the assignment `s1->len = 0`, the observed bounds for `s1->f` are `bounds(unknown)`. This results in an error when attempting to use `s1->f` in a dynamic bounds cast.

This PR uses an initial dynamic bounds cast so that the observed bounds of `s1->f` are depending on a temporary variable rather than `s1->len`, so that the assignment to `s1->len` does not impact the observed bounds of `s1->f`.

### DIFF
--- a/tests/dynamic_checking/bounds/bounds-safe-interfaces.c
+++ b/tests/dynamic_checking/bounds/bounds-safe-interfaces.c
@@ -210,15 +210,24 @@ void write_driver(int failure_point, int *a1 : count(10),
       break;
     // Vary structure lengths.
     case 13:
-      s1->len = 0, s1->f = _Dynamic_bounds_cast<_Array_ptr<int>>(s1->f, bounds(s1->f, s1->f + s1->len));
+      int temp0 = s1->len;
+      s1->f = _Dynamic_bounds_cast<_Array_ptr<int>>(s1->f, bounds(s1->f, s1->f + temp0)),
+        s1->len = 0,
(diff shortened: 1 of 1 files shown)


tests/dynamic_checking/bounds/bounds-safe-interfaces.c
```diff
@@ -210,15 +211,21 @@ void write_driver(int failure_point, int *a1 : count(10),
       break;
     // Vary structure lengths.
     case 13:
-      s1->len = 0, s1->f = _Dynamic_bounds_cast<_Array_ptr<int>>(s1->f, bounds(s1->f, s1->f + s1->len));
+      s1->f = _Dynamic_bounds_cast<_Array_ptr<int>>(s1->f, bounds(s1->f, s1->f + temp)),
+        s1->len = 0,
+        s1->f = _Dynamic_bounds_cast<_Array_ptr<int>>(s1->f, bounds(s1->f, s1->f + s1->len));
```
sulekhark, line 216: Instead of creating a `temp` variable, can the solution style adopted in PR #116 of `checkedc-llvm-test-suite` repo be adopted here and in all the scenarios below?

sulekhark (approved):
LGTM! Thank you!

mgrang (approved):
LGTM.

PR #505, opened 2024-09-16
Fixes for MacOS CI runs.

dtarditi:
This change fixes MacOS CI runs using GitHub's macos-latest runner.  They go with this [PR](https://github.com/checkedc/workflows/pull/2) in the workflows repo.  

Apple has made some changes to the header files for malloc and related functions to use variables with the name ptr. This conflicts with the macro in stdcheckedc.h defining ptr as _Ptr.  Remove the uses of stdcheckedc.h in files that fail on MacOS.


Machiry (approved):
Looks good.

